PR #40, opened 2022-07-20
Multi-batch support for text-generation

jeffra:
Adds single and multi-batch support for text-generation models


mii/grpc_related/proto/modelresponse.proto
```diff
@@ -41,15 +41,13 @@ message Value {
...
 
-// The request message containing the user's name.
 message SingleStringRequest {
-  string request = 1;
+  repeated string request = 1;
   map<string,Value> query_kwargs = 2;
 }
 
-// The response message containing the greetings
 message SingleStringReply {
```
mrwyattii, line 49: Perhaps we should rename these from `SingleString*` to `ListString*` or `MultiString*`?

mrwyattii (approved):
LGTM

PR #43, opened 2022-07-25
Bloom custom ckpt files, improved model path support, MII refactor, flake8, etc.

jeffra:
* Task specific request_dict parsing, ensures required keys are given for each task type.
* Support appending `model_path` and aml model path in aml deployments
* Pass checkpoint dict to ds-inference instead of writing to file
* Add flake8 pre-commit and associated fixes
* Big refactor of overall code structure

Requires DeepSpeed version that includes this: https://github.com/microsoft/DeepSpeed/pull/2132


mii/deployment.py
```diff
@@ -102,29 +64,36 @@ def deploy(task,
...
 
-    create_score_file(deployment_name,
-                      task,
-                      model,
-                      enable_deepspeed,
-                      enable_zero,
-                      ds_config,
-                      mii_config)
+    # In local deployments use default path if no model path set
+    if model_path is None and deployment_type == DeploymentType.LOCAL:
```
mrwyattii, line 75: Do we need a similar if statement for `model_path is None and deployment_type == DeploymentType.AML`?

mii/models/load_models.py
```diff
@@ -176,27 +29,30 @@ def load_models(task_name,
     #TODO: pass in mii_config to fetch dtype, training_mp_size, and other params
 
     checkpoint = None
+    checkpoint_path = None
```
mrwyattii, line 32: It appears this never gets used (why doesn't flake8 precommit catch this?)

mrwyattii (changes requested):
Looks good so far

mii/config.py
```diff
@@ -7,13 +7,25 @@ class MIIConfig(BaseModel):
...
+    @validator('checkpoint_dict')
+    def checkpoint_dict_valid(cls, value):
+        if not (isinstance(value, dict) or value is None):
+            raise ValueError(f"checkpoint_dict should be dict or None")
+        if isinstance(value, dict):
+            if value.get('base_dir', ''):
+                raise ValueError(
+                    "please unset 'base_dir' it will be set w.r.t. the deployment 'model_path'"
+                )
+        return value
```
mrwyattii, line 27:
```suggestion
        if 'base_dir' in value:
            raise ValueError(
                "please unset 'base_dir' it will be set w.r.t. the deployment 'model_path'"
            )
        return value
```

mrwyattii: We don't need to check for dict type here, pydantic will handle that for us

mii/utils.py
```diff
@@ -110,17 +110,23 @@ def check_if_task_and_model_is_valid(task, model_name):
...
-        return os.path.join(aml_model_dir, model_path)
+def full_model_path(model_path):
+    aml_model_dir = os.environ.get('AZUREML_MODEL_DIR', None)
+
+    if aml_model_dir:
+        # (potentially) append relative model_path w. aml path
+        assert os.path.isabs(aml_model_dir), f"AZUREML_MODEL_DIR={aml_model_dir} must be an absolute path"
+        if model_path:
+            assert not os.path.isabs(model_path), f"model_path={model_path} must be relative to append w. AML path"
+            return os.path.join(model_path, aml_model_dir)
```
mrwyattii, line 121:
```suggestion
            return os.path.join(aml_model_dir, model_path)
```

The `model_path` should be appended, not preprended - right?

mii/utils.py
```diff
@@ -110,17 +110,23 @@ def check_if_task_and_model_is_valid(task, model_name):
...
+        else:
+            return aml_model_dir
     else:
```
mrwyattii, line 124:
```suggestion
    elif model_path:
```

mii/utils.py
```diff
@@ -110,17 +110,23 @@ def check_if_task_and_model_is_valid(task, model_name):
...
-        return aml_model_dir
+        # no aml path is set
+        if model_path:
+            return model_path
+        else:
+            return mii.constants.MII_MODEL_PATH_DEFAULT
```
mrwyattii, line 129:
```suggestion
        return model_path
    else:
        return mii.constants.MII_MODEL_PATH_DEFAULT
```

mii/utils.py
mrwyattii: remove unnecessary extra if-else level

mii/utils.py
jeffra: Ohh great catch, yes you are right.

mrwyattii (approved):

PR #57, opened 2022-09-30
add support for txt2img task

jeffra:
Adds support for txt2img task in MII. This feature is dependent on the DS side PR: https://github.com/microsoft/DeepSpeed/pull/2381


mii/models/load_models.py
```diff
@@ -27,6 +27,8 @@ def load_models(task_name,
     mpu = None
     args = None
     training_mp_size = 1
+    kernel_inject = True
+    replace_pipeline_model = True
```
mrwyattii, line 31: This seems to be an unused variable

mrwyattii (approved):
LGTM - should we also refactor the other examples to include the example/query in a single script?

mii/models/load_models.py
jeffra: good catch, fixed :)

awan-10 (approved):
Looks good!

PR #59, opened 2022-10-07
various clean-up

jeffra:
* Exposes two ds-inference parameters: replace_with_kernel_inject and profile_model_time
* Allows float32/16 dtype names
* Adds opt support mapped to HF provider
* Collapse all DS kwargs into single dict
* Allow bloom-176b w. int8
* Iff DS supports enable_qkv_quantization set it to true for bloom-176b
* Update bloom-176b tokenizer to use padding side left and pad token
* Copied changes from #45 into this PR, they are not always needed but still a nice usability feature

### DIFF
--- a/mii/models/providers/llm.py
+++ b/mii/models/providers/llm.py
@@ -3,6 +3,7 @@ import torch
 import deepspeed
 from deepspeed.inference.engine import InferenceEngine
 from deepspeed import OnDevice
+from mii.utils import mii_cache_path
 
 from transformers import AutoModelForCausalLM, AutoTokenizer, AutoConfig
 from transformers.utils import WEIGHTS_NAME, WEIGHTS_INDEX_NAME, cached_path, hf_bucket_url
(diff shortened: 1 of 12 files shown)


mii/models/providers/llm.py
```diff
@@ -101,15 +103,32 @@ def create_checkpoint_dict(model_name, model_path, mii_config):
...
+
+    tokenizer = _attempt_load(AutoTokenizer.from_pretrained,
+                              model_name,
+                              cache_path,
+                              kwargs={"padding_side": 'left'})
+    tokenizer.pad_token = tokenizer.eos_token
+
+    config = _attempt_load(AutoConfig.from_pretrained, model_name, cache_path)
+
+    with OnDevice(dtype=torch.float16, device='meta', enabled=True):
```
mrwyattii, line 131: Should the dtype here be hard-coded? What about the bloom int8 model?

jeffra: Good question, this creates the `meta` tensors of the model that are eventually overwritten by the DS-inference checkpoint load. My understanding is we need ds-inference to do the fp16->int8 dtype conversion, but would be good to have @RezaYazdaniAminabadi confirm here.

RezaYazdaniAminabadi: I think it is fine to pass fp16 there since there is no real tensor created and we will inject whatever data required based on the configuration.

mrwyattii (approved):
LGTM

PR #61, opened 2022-10-07
Staging ammar (#3)

awan-10:
merge and cleanup

### DIFF
--- a/mii/grpc_related/modelresponse_server.py
+++ b/mii/grpc_related/modelresponse_server.py
@@ -28,17 +28,18 @@ class ModelResponse(modelresponse_pb2_grpc.ModelResponseServicer):
 
     def _get_model_time(self, model, sum_times=False):
         model_times = []
-        if hasattr(model, "model_times"):
+        # Only grab model times if profiling was enabled/exists
+        if getattr(model, "model_profile_enabled", False):
             model_times = model.model_times()
 
-        if len(model_times) == 1:
-            model_time = model_times[0]
-        elif len(model_times) > 1:
+        if len(model_times) > 0:
             if sum_times:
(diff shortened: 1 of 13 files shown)


mii/grpc_related/modelresponse_server.py
```diff
@@ -26,13 +26,29 @@ def _unpack_proto_query_kwargs(self, query_kwargs):
...
+        if len(model_times) == 1:
+            model_time = model_times[0]
+        elif len(model_times) > 1:
+            if sum_times:
+                model_time = sum(model_times)
+            else:
+                model_time = model_times[-1]
+        else:
+            model_time = -1
+        return model_time
```
mrwyattii, line 43: Maybe add comments to help understand this. Why would model_times sometimes have len==1 and sometimes have multiple entries? Also what is the significance of returning -1 when len(model_times)==0?

mii/models/providers/llm.py
```diff
@@ -101,15 +103,32 @@ def create_checkpoint_dict(model_name, model_path, mii_config):
...
+
+    tokenizer = _attempt_load(AutoTokenizer.from_pretrained,
+                              model_name,
+                              cache_path,
+                              kwargs={"padding_side": 'left'})
+    tokenizer.pad_token = tokenizer.eos_token
+
+    config = _attempt_load(AutoConfig.from_pretrained, model_name, cache_path)
+
+    with OnDevice(dtype=torch.float16, device='meta', enabled=True):
```
mrwyattii, line 131: dtype here should probably follow the model dtype and not be hardcoded to float16

jeffra: moved to other PR

PR #70, opened 2022-10-15
fix typo

TahaBinhuraib:


TahaBinhuraib:
should be `f"> --------- MII Settings: {ds_optimize}`
 was resulting in an error when importing mii
 should be a simple fix

mustafakemalelma:
Please merge this.

mrwyattii:
@mustafakemalelma & @TahaBinhuraib what version of Python are you using that this line is causing an error? This works for me with Python 3.8 (but I'm not exactly sure when support for formatting like `f"{my_var=}"` was added)

mrwyattii (approved):

PR #79, opened 2022-10-28
refactor(example): conversational-query

Thytu:
* replace hard-coded `conversation_id` by a variable one (helps to understand that `3` is not a magic number but one to be set
* renamed `str` by `text` (has `str` is a python keyword ands thus should not be overwrited)
* un-comment commented code (show the usage end2end)


Thytu:
@microsoft-github-policy-service agree [company="GladIA"]

Thytu:
@microsoft-github-policy-service agree company="GladIA"

examples/local/conversational-query-example.py
```diff
@@ -1,30 +1,33 @@
...
 print(f"Querying {name}...")
 
-str = "DeepSpeed is the greatest"
+text = "DeepSpeed is the greatest"
 
 generator = mii.mii_query_handle(name + "_deployment")
 result = generator.query({
     'text': str,
-    'conversation_id': 3,
+    'conversation_id': time(),
```
mrwyattii, line 16: I think the better solution would be to:
- use a uuid instead of current time
- allow this to be an optional parameter ([hugginface.Conversation](https://huggingface.co/transformers/v4.11.3/main_classes/pipelines.html#transformers.Conversation) will auto-generate a uuid if no value is provided)

@jeffra thoughts?

Thytu: Good point, will update the code to use `UUID4` 👌

Thytu: About making `'conversation_id'` as an optional argument. I would rather do it in another PR as it's not "example" related but rather "core" related.

mrwyattii: > @mrwyattii It seems that with the current code implementation, `conversation_id` can only be an `int`.
> 
> Otherwise it will encounter the following error message : `TypeError: 'UUID' object cannot be interpreted as an integer`
> 
> Raise by : `mii/server_client.py", line 291, in _request_async_response`
> 
> It don't think that the desired behaviour as you stated, [transformers.Conversation](https://huggingface.co/transformers/v4.11.3/main_classes/pipelines.html#transformers.Conversation) expect `conversation_id` to be either type of `uuid.UUID` or `None`.
> 
> I suggest to keep forward with the `time` option for the moment until being able to provide an `UUID` type. @mrwyattii is it fine to you?
> 
> Side note : Actually you can give it anything as argument it will still works fine.
> 
> i.e :
> 
> ```python
> conversation = Conversation(
>     text="Going to the movies tonight - any suggestions?",
>     conversation_id=lambda X : X # <-- Here
> )
> 
> from transformers import Conversation, pipeline
> 
> conversational_pipeline = pipeline("conversational")
> 
> result = conversational_pipeline([conversation])
> ```

I think the problem is in our prototext, where we expect the value passed in to be an `int`. This clearly doesn't align with `transformers.Conversation`. @jeffra and I were discussing refactoring the inputs for this particular task so that they are passed like other kwargs to `Conversation`. This will probably be in a separate PR. @jeffra how do you feel about using `time()` or perhaps defining a separate variable `conv_id = 3` to demonstrate that this is not a magic number in the example?

Thytu: Using `conv_id = 3` would be better than using time (my2c).

I would just suggest setting `conv_id` as `0`, again to demonstrate that `3` is not a magic number.

Thytu: > @jeffra and I were discussing refactoring the inputs for this particular task so that they are passed like other kwargs to Conversation

If I can help in any way, you be glad too ;)

jqueguiner:
@Thytu to pass the yapf failed format test :
https://github.com/microsoft/DeepSpeed-MII/blob/main/CONTRIBUTING.md#prerequisites

Thytu:
UT fails due to `huggingface_hub.utils._errors.HfHubHTTPError: 500 Server Error: Internal Server Error for url: https://huggingface.co/api/models (Request ID: TQhG-2CQiatqo3lBxTT8u`

Not on my side

mrwyattii:
> UT fails due to `huggingface_hub.utils._errors.HfHubHTTPError: 500 Server Error: Internal Server Error for url: https://huggingface.co/api/models (Request ID: TQhG-2CQiatqo3lBxTT8u`
> 
> Not on my side

I think this is resolved on the HF side now, but we are failing UT for MII due to another change on the DeepSpeed side. Will be resolved by https://github.com/microsoft/DeepSpeed/pull/2533

examples/local/conversational-query-example.py
Thytu: updated by [68d7d0e](https://github.com/microsoft/DeepSpeed-MII/pull/79/commits/68d7d0ef6a32c7a968aec8e1642983ffa9e09406)

Thytu:
> > UT fails due to `huggingface_hub.utils._errors.HfHubHTTPError: 500 Server Error: Internal Server Error for url: https://huggingface.co/api/models (Request ID: TQhG-2CQiatqo3lBxTT8u`
> > Not on my side
> 
> I think this is resolved on the HF side now, but we are failing UT for MII due to another change on the DeepSpeed side. Will be resolved by [microsoft/DeepSpeed#2533](https://github.com/microsoft/DeepSpeed/pull/2533)

Fixed, well done

mrwyattii (approved):

PR #108, opened 2022-11-22
Use new DS-Inference config API

mrwyattii:


jeffra (approved):

jeffra:
This requires deepspeed 0.7.6+, should we bump the min deepspeed version in the mii requirements?

PR #121, opened 2022-12-19
Free torch cache after loading model

mrwyattii:
`nvidia-smi` will report the wrong amount of memory used due to cached memory from loading / splitting a model across GPUs. Freeing this cached memory makes it so the correct amount of GPU memory usage is shown.


jeffra (approved):
Seems fine to me. Perhaps we should also call `deepspeed.runtime.utils.see_memory_usage()` here? nvidia-smi is pretty misleading even with the empty cache call.

https://github.com/microsoft/DeepSpeed/blob/6273dffc2f192275a08268b683c309a328b52191/deepspeed/runtime/utils.py#L821

Issue #106, opened 2022-11-21
Support for FLAN-T5


jihan-yin:
I saw that T5 wasn't in the list of supported huggingface transformers models. Are there plans / ETA for when the T5 family would be added? FLAN-T5 is a very strong llm for zero/fewshot instruction prompting. I am currently building out a hacky implementation for hosting with deepspeed-inference, but having it natively supported in deepspeed-mii would be ideal.

jeffra:
Also keep an eye on this PR, it’s currently a work in progress for better T5 support: https://github.com/microsoft/DeepSpeed/pull/2451

PR #124, opened 2022-12-25
Support text2text-generation pipelines and T5 models in particular

philhchen:
**Summary**
Support the T5 family of models on Huggingface. Additionally, since the T5 models use the `text2text-generation` pipelines on Huggingface, this PR adds support for `text2text-generation` pipelines. Because the `text2text-generation` pipelines are conceptually very similar to `text-generation` pipelines, this PR uses the same `GeneratorReply` stub for model serving.

**Linked Issues**
Addresses https://github.com/microsoft/DeepSpeed-MII/issues/106

**Usage guide**
```commandline
❯ python examples/local/text-generation-flan-example.py
❯ python examples/local/text-generation-flan-query-example.py
Querying flan...
response: "DeepSpeed is the fastest and most reliable speed of the ethernet."
time_taken: 0.546296835
model_time_taken: -1

time_taken: 0.5462968349456787
```

### DIFF
--- a/examples/local/text-generation-flan-example.py
+++ b/examples/local/text-generation-flan-example.py
@@ -1,0 +1,7 @@
+import mii
+
+mii_configs = {"tensor_parallel": 1, "dtype": "fp16"}
+mii.deploy(task='text2text-generation',
+           model="google/flan-t5-small",
+           deployment_name="flan_deployment",
+           mii_config=mii_configs)
--- a/examples/local/text-generation-flan-query-example.py
+++ b/examples/local/text-generation-flan-query-example.py
@@ -1,0 +1,16 @@
+import mii
+
+print(f"Querying flan...")
+
+str = "Tell me why DeepSpeed is the greatest."
+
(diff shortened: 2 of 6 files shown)


philhchen:
@microsoft-github-policy-service agree [company=""]

philhchen:
@microsoft-github-policy-service agree

jeffra:
Thank you @philhchen for your PR and for using MII! DeepSpeed inference does not currently support injecting kernels for T5. However, there is ongoing work to support this type of model, feel free to watch this PR: https://github.com/microsoft/DeepSpeed/pull/2451

We could however add support in MII for models that don't have kernel-level acceleration. Currently kernel-injection is on by default: https://github.com/microsoft/DeepSpeed-MII/blob/4040daecc4185e591d25cffc07b9cfd6de9f4fb7/mii/config.py#L44

We could add some code like yours that allows models like T5 but forces the user to set `replace_with_kernel_inject = False` in the mii config. Would this be useful?

loadams:
Closing as stale and due to merge conflicts.

PR #129, opened 2022-12-30
Add max_tokens option to mii_config

mallorbc:
Currently, there is no way to specify what the max amount of tokens one can use should be.  For GPTJ, this defaults it to be 1024 tokens which is half of its max of 2048.

Through this PR, one can now specify the max number of tokens through something like the following:

``` python
mii_configs = {"tensor_parallel": 1, "dtype": "fp16",max_tokens":2048}
mii.deploy(task="text-generation",
           model="EleutherAI/gpt-j-6B",
           deployment_name="deployment",
           mii_config=mii_configs)
```

I have tested this for GPTJ and it works.


mallorbc:
@microsoft-github-policy-service agree [company="Brillibits LLC"]

mallorbc:
@microsoft-github-policy-service agree company="Brillibits LLC"

mallorbc:
@microsoft-github-policy-service agree

jeffra:
Hi @mallorbc, thank you for your PR and for trying MII. You can set generation arguments like this in the query itself after deploy, here's an example below. Would this work for your use case?

```python
import mii
generator = mii.mii_query_handle("bloom560m_deployment")
result = generator.query({"query": ["DeepSpeed is", "Seattle is"]}, do_sample=True, max_new_tokens=30)
print(result)
```

mallorbc:
> Hi @mallorbc, thank you for your PR and for trying MII. You can set generation arguments like this in the query itself after deploy, here's an example below. Would this work for your use case?
> 
> ```python
> import mii
> generator = mii.mii_query_handle("bloom560m_deployment")
> result = generator.query({"query": ["DeepSpeed is", "Seattle is"]}, do_sample=True, max_new_tokens=30)
> print(result)
> ```

Hey, thanks for the response.  My understanding is that does not work due to the way DeepSpeed works.  When to start the DeepSpeed engine, you tell it what is the max number of possible tokens it could need to generate.  It then allocates memory for that during generation.

To demonstrate the need for this, try generating more than 1024 tokens without my PR.  You will get junk after the 1024 tokens.  With the PR, and if configured to use 2048 tokens, it will use more memory, but the generated tokens will not be junk.

cmikeh2:
> > Hi @mallorbc, thank you for your PR and for trying MII. You can set generation arguments like this in the query itself after deploy, here's an example below. Would this work for your use case?
> > ```python
> > import mii
> > generator = mii.mii_query_handle("bloom560m_deployment")
> > result = generator.query({"query": ["DeepSpeed is", "Seattle is"]}, do_sample=True, max_new_tokens=30)
> > print(result)
> > ```
> 
> Hey, thanks for the response. My understanding is that does not work due to the way DeepSpeed works. When to start the DeepSpeed engine, you tell it what is the max number of possible tokens it could need to generate. It then allocates memory for that during generation.
> 
> To demonstrate the need for this, try generating more than 1024 tokens without my PR. You will get junk after the 1024 tokens. With the PR, and if configured to use 2048 tokens, it will use more memory, but the generated tokens will not be junk.

This is accurate, since the `max_out_tokens` of the DeepSpeed configuration (https://github.com/microsoft/DeepSpeed/blob/a091bc223c01e94448f443456a6c15684644b966/deepspeed/inference/config.py#L240) is used to determine the scratchpad allocation, while the `max_out_tokens` on the generation call is only used by HuggingFace to control stopping conditions. Long term, a solution would be to watch that argument to generate and update the allocation code based on the generation argument, but there's not infrastructure for that yet.

jeffra:
Ohh I see, totally misunderstood the intent here. If that's the intent then we need to be sure to pass mii-config `max_out_tokens` value to `inf_config` that is passed to deepspeed init here: https://github.com/microsoft/DeepSpeed-MII/blob/4040daecc4185e591d25cffc07b9cfd6de9f4fb7/mii/models/load_models.py#L83-L86

mallorbc:
> Ohh I see, totally misunderstood the intent here. If that's the intent then we need to be sure to pass mii-config `max_out_tokens` value to `inf_config` that is passed to deepspeed init here:
> 
> https://github.com/microsoft/DeepSpeed-MII/blob/4040daecc4185e591d25cffc07b9cfd6de9f4fb7/mii/models/load_models.py#L83-L86

The PR does copy the value in the mii-config and sets it in the inf_config which is then used by DeepSpeed

jeffra:
Oh I see it now, `max_tokens` and `max_out_tokens` are both supported in ds-inference. Cool, we're good then :) Sorry for my confusion. LGTM then!

jeffra (approved):

mrwyattii:
@mallorbc Can you run formatting on this branch and update so we can get this merged? Thanks!

https://www.deepspeed.ai/contributing/#prerequisites

mallorbc:
> @mallorbc Can you run formatting on this branch and update so we can get this merged? Thanks!
> 
> https://www.deepspeed.ai/contributing/#prerequisites

I will do so as soon as I can.  Hopefull today or tomorrow.  Thanks!

mallorbc:
@mrwyattii sorry for the delay.  Forgot about this.  I think it's good now though.  I ran it through formatting and merged with updates

mrwyattii:
Merging this, but we should think about how to refactor MII to allow users to specify other DS-inference kwargs and not just the `max_tokens`.
@jeffra @tohtana

PR #138, opened 2023-01-27
Decouple conversions for gRPC from server/client code 

tohtana:
This PR aims to simplify the code of server and client by decoupling conversions for gRPC.

Currently, the following conversions are implemented for each task in the server or client code, and this can make the code difficult to maintain.

- client API inputs -> gRPC request -> pipeline inputs
- pipeline outputs -> gRPC response -> client API 

The changes will involve consolidating the conversions into the new table in `method_table.py`.

The future extensions (e.g. load balancing, dynamic batching) will benefit from separating task-specific logic from general functions of the server and client code. 
In addition, developers can also easily know from the table how to add a new task type.

### DIFF
--- a/mii/grpc_related/modelresponse_server.py
+++ b/mii/grpc_related/modelresponse_server.py
@@ -5,18 +5,18 @@ from concurrent import futures
 import logging
 
 import grpc
-from .proto import modelresponse_pb2
 from .proto import modelresponse_pb2_grpc
 import sys
 import time
 
-from transformers import Conversation
 from mii.constants import GRPC_MAX_MSG_SIZE
+from mii.method_table import GRPC_METHOD_TABLE
 
 
 class ModelResponse(modelresponse_pb2_grpc.ModelResponseServicer):
     def __init__(self, inference_pipeline):
         self.inference_pipeline = inference_pipeline
+        self.method_name_to_task = {m["method"]: t for t, m in GRPC_METHOD_TABLE.items()}
 
     def _unpack_proto_query_kwargs(self, query_kwargs):
         query_kwargs = {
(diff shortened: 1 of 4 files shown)


mii/grpc_related/modelresponse_server.py
```diff
@@ -44,120 +44,48 @@ def _get_model_time(self, model, sum_times=False):
...
+        if method_name not in self.method_name_to_task:
+            raise ValueError(f"unknown method: {method_name}")
+
+        task = self.method_name_to_task[method_name]
+        if task not in GRPC_METHOD_TABLE:
+            raise ValueError(f"unknown task: {task}")
 
-        # unpack grpc list into py-list
-        request = [r for r in request.request]
+        conv_funcs = GRPC_METHOD_TABLE[task]
```
mrwyattii, line 55: Can we get a better name for this variable? I think `conv` means `conversion`, but it's unclear.

mii/method_table.py
```diff
@@ -0,0 +1,158 @@
...
+    return modelresponse_pb2.MultiStringReply(response=text_responses,
+                                              time_taken=time_taken,
+                                              model_time_taken=model_time_taken)
+
+
+def question_answering_unpack_request_from_proto(request):
+    kwargs = unpack_proto_query_kwargs(request.query_kwargs)
+    kwargs["question"] = request.question
+    kwargs["context"] = request.context
+    return (), kwargs
```
mrwyattii, line 55:
```suggestion
    args = ()
    return args, kwargs
```

I think this makes it more clear what we are returning from these functions. Type hints for the return types (or even just comments) may also be useful for future developers adding new tasks.

mii/method_table.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def conversational_unpack_request_from_proto(request):
+    kwargs = unpack_proto_query_kwargs(request.query_kwargs)
+    conv = Conversation(text=request.text,
+                        conversation_id=request.conversation_id,
+                        past_user_inputs=request.past_user_inputs,
+                        generated_responses=request.generated_responses,
+                        **kwargs)
+    return (conv, ), {}
```
mrwyattii, line 65:
```suggestion
    args = (conv, )
    kwargs = {}
    return args, kwargs
```

mii/grpc_related/modelresponse_server.py
```diff
@@ -44,120 +44,48 @@ def _get_model_time(self, model, sum_times=False):
...
 
-        model_time = self._get_model_time(self.inference_pipeline.model, sum_times=True)
+        return conv_funcs["pack_response_to_proto"](response, end - start, model_time)
 
-        val = modelresponse_pb2.MultiStringReply(response=text_responses,
-                                                 time_taken=end - start,
-                                                 model_time_taken=model_time)
-        return val
+    def GeneratorReply(self, request, context):
+        return self._run_inference("GeneratorReply", request)
```
mrwyattii, line 70: This refactor has me wondering if these functions are really necessary. For example, what if we just passed the `TASK_NAME` value to `_run_inference` from here? https://github.com/microsoft/DeepSpeed-MII/blob/b86e0ce9dbffe1ca75f7f9bd44a5f96a3999542e/mii/server_client.py#L267

Additionally, I think the way we currently pass `request_dict` and `query_kwargs` to [query](https://github.com/microsoft/DeepSpeed-MII/blob/b86e0ce9dbffe1ca75f7f9bd44a5f96a3999542e/mii/server_client.py#L349) could be simplified to `*args, **kwargs`. These args and kwargs would then be passed all the way to the pipeline object. This way users would run inference on MII with the same args and kwargs they would use with a `transformers.pipeline` object.

mii/method_table.py
```diff
@@ -0,0 +1,158 @@
...
+        "ConversationalReply",
+        "pack_request_to_proto":
+        lambda request_dict,
+        **query_kwargs: modelresponse_pb2.ConversationRequest(
+            text=request_dict['text'],
+            conversation_id=request_dict['conversation_id']
+            if 'conversation_id' in request_dict else None,
+            past_user_inputs=request_dict['past_user_inputs'],
+            generated_responses=request_dict['generated_responses'],
+            query_kwargs=kwarg_dict_to_proto(query_kwargs)),
```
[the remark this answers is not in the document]
tohtana: I'll change it as you suggested.

mii/grpc_related/modelresponse_server.py
tohtana: What about simply `conversions`?

mii/method_table.py
tohtana: I will change it as you suggested. For type hints, let me discuss in the response to other comments.

mii/grpc_related/modelresponse_server.py
tohtana: I understand your point and agree that the current change looks redundant.

My concern is about serialization. If you use the same gPRC API for all tasks, the message needs to be able to contain any type of request and such a message is not very efficient. It seems natural that we have different message schemas for texts and images. This is also related to type hints. Once you use the same API for all tasks, you cannot give a specific data type to the request/response to a task type.

To reduce the redundancy of the method names, another approach could be to set a hook on the superclass, which is generated from the proto, to call `_run_inference` based on the method name. But I'm also wondering if it might be too complicated to use some reflection-like approach for this refactoring work.

I would be happy if you could give me any feedback.

tohtana: We will offer APIs that accept the same inputs as those to `transformers.pipeline` in the future, but not include the change in this PR.

mii/grpc_related/modelresponse_server.py
mrwyattii: `conversions` or `conversions_funcs` would work

mrwyattii (approved):
LGTM

jeffra (approved):

jeffra:
Thanks @tohtana!

PR #142, opened 2023-02-01
Separate server and client

tohtana:
Currently `MIIServerClient` works as both server and client, but they do not share much of the code.
This PR aims to split it to two different classes: MIIServer and MIIClient.
We can keep the code for each small and increase the maintainability with this change.

Regarding this change, this PR also removes the flag `use_grpc_server`.
Now gRPC is used for both LOCAL and AML settings and the flag is always set to True in `score.py`.
We can drastically simplify many conditional branches in server and client by removing the flag.

### DIFF
--- a/mii/utils.py
+++ b/mii/utils.py
@@ -5,6 +5,7 @@ import sys
 import os
 import logging
 import importlib
+import torch
 import mii
 
 from huggingface_hub import HfApi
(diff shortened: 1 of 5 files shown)


mii/utils.py
```diff
@@ -199,6 +200,18 @@ def extract_query_dict(task, request_dict):
...
+def get_num_gpus(mii_configs):
+    def get_tensor_parallel_gpus(mii_configs):
+        num_gpus = mii_configs.tensor_parallel
+
+        assert torch.cuda.device_count(
+        ) >= num_gpus, f"Available GPU count: {torch.cuda.device_count()} does not meet the required gpu count: {num_gpus}"
+        return num_gpus
+
+    # Only Tensor Parallelism supported for now
+    return get_tensor_parallel_gpus(mii_configs)
```
mrwyattii, line 212:
```suggestion
def get_num_gpus(mii_configs):
    num_gpus = mii_configs.tensor_parallel

    assert torch.cuda.device_count(
    ) >= num_gpus, f"Available GPU count: {torch.cuda.device_count()} does not meet the required gpu count: {num_gpus}"
    return num_gpus
```
I know this was how the method was defined in the server_client code, but can we simply this?

jeffra (approved):

PR #147, opened 2023-02-08
Load balancing and multiple replicas

tohtana:
This PR enables users to launch multiple replicas of MII server instances and a load balancer.
Here is an example of the configuration.

```python
mii_configs = {"tensor_parallel": tensor_parallel,
                   "enable_load_balancing": True,
                   "replica_num": replica_num,
                   "hostfile": hostfile
                   }
mii.deploy(task=task,
               model=model_name,
               deployment_name=deployment_name,
               mii_config=mii_configs,
               deployment_type=DeploymentType.LOCAL)
```

Now we have three new configuration items: ``enable_load_balancing``, ``replica_num``, and ``hostfile``.

The behaviors of MII do not change when ``enable_load_balancing==False``.

When ``enable_load_balancing==True``, `mii.deploy()` launches as many instances as `replica_num` and the load balancer server.
Note that each instance consists of ``tensor_parallel`` servers.
The client sends gRPC request to the load balancer instead of MII servers. The load balancer forwards the request to MII servers. Currently the load balancer implements the simple round robin algorithm.

`hostfile` is the path to [hostfile](https://www.deepspeed.ai/getting-started/#resource-configuration-multi-node) used by DeepSpeed's launcher. When `hostfile` is not specified, DeepSpeed-MII uses the default path `/job/hostfile`, which is defined for DeepSpeed.

lb_overview

These are the notable changes in the code.
- Use deepspeed to launch servers on multiple nodes
- Now we have two client classes: MIIClient and MIITensorParallelClient. The former is used to call load balancer and the latter is used to call MII tensor parallel servers in parallel. ``mii.mii_query_handle()`` chooses one according to the settings in `score.py`.
- Terminate servers via a gRPC call because the servers can run on different nodes and we cannot directly kill the processes.

One concern is how to test. I tested this feature on the single node with a dummy hostfile like
```
worker-1 slots=8
worker-2 slots=8
```
but I needed to edit hosts file on the server. MII also needs to access the target server without any user interaction.

The changes in this PR are not very small and there are many other approaches for this feature.
I would appreciate if you could give me any feedback.


mii/deployment.py
```diff
@@ -98,6 +100,23 @@ def deploy(task,
...
+                                       mii_config.tensor_parallel,
+                                       mii_config.replica_num)
+    replica_deployment = []
+    for i, (host, gpu_indices) in enumerate(replica_pool):
+        grpc_ports = []
+        # Reserver port for a LB proxy when replication is enabled
+        port_offset = 1 if mii_config.replica_num > 1 else 0
+        for j in range(mii_config.tensor_parallel):
+            grpc_ports.append(mii_config.port_number + i * mii_config.tensor_parallel +
+                              j + port_offset)
```
mrwyattii, line 114: This could be simplified:
```python
port_offset = 1 if mii_config.replica_num > 1 else 0
base_port = mii_config.port_number + i * mii_config.tensor_parallel + port_offset
grpc_ports = range(base_port, base_port+mii_config.tensor_parallel)

mrwyattii (approved):

jeffra (approved):

PR #154, opened 2023-03-02
RESTful API support

tohtana:
This PR enables users to call MII servers via RESTful APIs.

You can enable this feature by adding `restful_api_port` in MIIConfig.
The launcher starts the gateway server that converts a RESTful API request to gRPC request.
Then the gateway server forwards the request to the load balancer.
Thus, `enable_load_balancing` must be `True` when you enable this feature. You can set 1 to `replica_num` if you don't need load balancing.

An example of the configuration is:
```python
mii_configs = {
...
               "enable_load_balancing": True,
               "restful_api_port": PORT_FOR_RESTFUL_API,
...
 }
```

The endpoint of RESTful APIs are set at `http://localhost:[PORT_FOR_RESTFUL_API]/mii/[TASK_NAME]`.

You can call the API using any other HTTP client. An example of using curl is (assume `restful_api_port` is 28080):
```bash
$ curl --header "Content-Type: application/json" --request POST  -d '{"request": {"query": ["Seattle is", "Bellevue is", "Redmond is"]}, "kwargs": {"do_sample": false, "max_new_tokens": 100}}' http://localhost:28080/mii/text-generation
```

The code below is an example using Python.
```python
import requests
import json

# text_generation
url = 'http://localhost:28080/mii/text-generation'
params = {"request": {"query": ["Seattle is", "Bellevue is", "Redmond is"]},
          "kwargs": {"do_sample": False, "max_new_tokens": 100}}

json_params = json.dumps(params)
response = requests.post(url, data=json_params, headers={
                         "Content-Type": "application/json"})
print(response.json())
```

I think the following points should be discussed regarding this feature (These should be filed as separate issues)
- Any users can terminate servers by calling `terminate` API. This is not desirable behavior for most use cases. One possible solution is to generate or set a secret key when starting servers and send it as an argument of `terminate`
- We could remove `enable_load_balancing` option if the overhead of the load balancer is negligible. This will allow us to remove many conditional branches (`if args.enable_load_balancing: ...`). (proposed by @mrwyattii)
- `MIIClient` usually returns results in ProtocolBuffers and the RESTful gateway coverts the result to JSON. However, MIIClient returns a non-ProtocolBuffers value only for text-to-image and the RESTful API call fails. We could add postprocess for `text-to-image` but it is better to return results in a consistent format for all task types.

I can draft the document of load balancing and this feature if the design is approved.


mii/config.py
```diff
@@ -49,6 +49,7 @@ class MIIConfig(BaseModel):
     max_tokens: int = 1024
     enable_load_balancing: bool = False
     replica_num: int = 1
+    restful_api_port: int = 0
```
mrwyattii, line 52: This should probably be a reasonable default value. Port 0 would be the wildcard port

mii/grpc_related/restful_gateway.py
```diff
@@ -0,0 +1,62 @@
...
+import mii
+from flask import Flask, request
+from flask_restful import Resource, Api
+from werkzeug.serving import make_server
+from mii.constants import RESTFUL_GATEWAY_SHUTDOWN_TIMEOUT, RESTFUL_API_PATH
+from google.protobuf.json_format import MessageToJson
+
+
+def shutdown(thread):
+    time.sleep(RESTFUL_GATEWAY_SHUTDOWN_TIMEOUT)
```
mrwyattii, line 12: What purpose does this `sleep` serve when we terminate?

mii/grpc_related/restful_gateway.py
```diff
@@ -0,0 +1,62 @@
...
+    app = Flask("RestfulGateway")
+
+    @app.route("/terminate", methods=['GET'])
+    def terminate():
+        # Need to shutdown *after* completing the request
+        threading.Thread(target=shutdown, args=(server_thread, )).start()
+        return "Shutting down RESTful API gateway server"
+
+    api = Api(app)
+    path = '/{}/{}'.format(RESTFUL_API_PATH, task)
```
mrwyattii, line 39: I think it would be better to use the `deployment_name` rather than the `task` so we can have 2 MII servers running with the same task.

mii/server.py
```diff
@@ -318,6 +318,17 @@ def _initialize_service(self,
                                            mii_configs,
                                            lb_config))
 
+            if mii_configs.restful_api_port > 0:
```
mrwyattii, line 321: What if we used a `mii_configs.enable_restful_api` here instead of checking if the port is greater than 0? (and we could set a default port number for the restful api)

mii/config.py
mrwyattii: Also, if `load_balancing` has to be enabled for the restful api, would it make sense to automatically enable it when the restful api is enabled?

tohtana: Okay, I will add `enable_restful_api` to MIIConfig and set a reasonable number (e.g. 51080) as the default port.
I will also automatically enable `load_balancing` when `enable_restful_api` is True.

mii/grpc_related/restful_gateway.py
tohtana: The server would fail to return the response of `terminate` request if the server process finishes soon. This sleep delays the termination of the server process.

mii/grpc_related/restful_gateway.py
tohtana: I agree. I will change as proposed.

mii/server.py
tohtana: I will change as proposed (See also the above comment)

mii/config.py
```diff
@@ -84,6 +86,12 @@ def checkpoint_dict_valid(cls, value):
                 raise ValueError(f"Missing key={k} in checkpoint_dict")
         return value
 
+    @validator('enable_load_balancing')
+    def enable_load_balancing_valid(cls, field_value, values):
+        if values["enable_restful_api"]:
+            field_value = True
+        return field_value
+
```
mrwyattii, line 94:
```suggestion
    @root_validator
    def auto_enable_load_balancing(cls, values):
        if values["enable_restful_api"]:
            values["enable_load_balancing"] = True
        return values
```
If we make this a root validator, the ordering of the fields won't matter. With the current validator, if a future PR places the `enable_load_balancing` above `enable_restful_api` we would get the wrong behavior.

tohtana: Thank you for the proposal! I'm learning pydantic.

tests/test_local_deployment.py
```diff
@@ -215,6 +232,35 @@ def test_load_balancing(local_deployment, query):
...
+    for _ in range(2):
+        result = generator.query(query)
+
+    url = f'http://localhost:{restful_api_port}/mii/{local_deployment.deployment_name}'
+    params = {"request": query}
+    json_params = json.dumps(params)
+    result = requests.post(url,
+                           data=json_params,
+                           headers={"Content-Type": "application/json"})
+    assert result
```
mrwyattii, line 261: We probably want to assert that the result contains a response - I think this would pass on an error if the `requests.post` returned an error message.

mrwyattii (approved):

mrwyattii (approved):

jeffra (approved):

PR #177, opened 2023-04-28
Add session to enable multi-turn conversation

tohtana:
This PR adds a feature to handle multi-turn conversation scenarios.

We will have the following changes in MII's client APIs.

- `create_session(session_id)` / `destroy_session(session_id) are added to manage sessions
- kwarg `session_id` is added to `query()` to specify the session of the query

```python
import mii

generator = mii.mii_query_handle("DEPLOYMENT_NAME")
queries = ["Hello", "How are you doing?"]

session_id = "SESSION_NAME"
generator.create_session(session_id)

query_template = "Human: {}\n Assistant: "

for q in queries:
    result = generator.query({"query": query_template.format(q)}, session_id=session_id, max_new_tokens=128)
    print(result.response)

generator.destroy_session(session_id)
```

An example result is:
```bash
[' Yes, it is a good morning.  Thank you so much for contacting me.  I hope you are having a great day!</s>']
[' Yes, I am doing well.  Thank you so much for asking.  I hope you are having a great day as well.</s>']
```

The session ID is used for the following purposes:
- The load balancer always chooses the same replica when the session ID is same
- We can implement preprocess/postprocess using session ID for each task type. The preprocess/postprocess can store the context of the session.

This PR implements a preprocess and a postprocess for `text-generation`. They save previous inputs/outputs as a context and concatenate the previous context and a new input (In future, this part should be replaced with a more efficient approach using KV cache).
As defined in the preprocess, you can pass only one input (i.e. batch size = 1) with a session ID.

```python
def text_generation_preprocess_session(session_id, session_context, args, kwargs):
    if session_id not in session_context:
        raise ValueError(f"session {session_id} does not exist")
    if session_context[session_id] is None:
        session_context[session_id] = ""
    if len(args[0]) != 1:
        raise ValueError(f"You can pass only one prompt with a session_id")

    args = ([session_context[session_id] + args[0][0]], )
    return args, kwargs

def text_generation_postprocess_session(session_id,
                                        session_context,
                                        args,
                                        kwargs,
                                        response):
    generated_text = response[0][0]["generated_text"]
    session_context[session_id] = generated_text
    response[0][0]["generated_text"] = generated_text[len(args[0][0]):]
    return response

...

GRPC_METHOD_TABLE = {
    Tasks.TEXT_GENERATION: {
        "method": "GeneratorReply",
        "pack_request_to_proto": multi_string_request_to_proto,
        "unpack_request_from_proto": proto_request_to_list,
        "pack_response_to_proto": text_generation_pack_response_to_proto,
        "preprocess_session": text_generation_preprocess_session,
        "postprocess_session": text_generation_postprocess_session
    },
...
```


samyam (approved):
@tohtana after the readme file has been updated with the conversational examples we discussed, the PR looks all good to me.

mrwyattii (approved):

PR #190, opened 2023-05-22
Fix kwargs passing for `conversational` tasks

sakoush:
This PR fixes extra kwargs to `query` being passed to the underlying `.generate` method in the case of conversational tasks. Previously these kwargs were wired to a wrong object and causing the workload to fail.

Instead `conversational_unpack_request_from_proto` just returns `kwargs` from query proto to the caller, which will eventually be passed to `inference_pipeline`.

Updated `test_single_GPU` to cover this specific scenario as well.

### DIFF
--- a/mii/method_table.py
+++ b/mii/method_table.py
@@ -104,10 +104,8 @@ def conversational_unpack_request_from_proto(request):
     conv = Conversation(text=request.text,
                         conversation_id=request.conversation_id,
                         past_user_inputs=request.past_user_inputs,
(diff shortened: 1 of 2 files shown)


sakoush:
@microsoft-github-policy-service agree company="Seldon"

srsaggam:
This issue should be fixed by #211 @mrwyattii FYI

mrwyattii:
Fixed in #225 / #211. Closing

PR #197, opened 2023-05-31
Add Non-persistent deployment type

TosinSeg:
Add Non-persistent deployment type


tests/test_non_persistent_deployment.py
mrwyattii, line 1: We don't want to copy all the fixtures and tests from the other unit test file. We should re-use as many of the functions from the other file as possible (and perhaps move the fixtures to a new fixtures-only file).

mrwyattii (changes requested):

mii/client.py
```diff
@@ -34,14 +35,23 @@ def mii_query_handle(deployment_name):
...
     """
-    task_name, mii_configs = _get_deployment_info(deployment_name)
-    if mii_configs.enable_load_balancing:
-        return MIIClient(task_name, "localhost", mii_configs.port_number)
+    if deployment_name in mii.non_persistent_models:
+        assert deployment_name in mii.non_persistent_models, f"Could not find '{deployment_name}'"
+        inference_pipeline, task = mii.non_persistent_models[deployment_name]
+        return MIINonPersistentClient(task, deployment_name)
+
     else:
```
mrwyattii, line 43: We can remove this `else` and avoid an extra level of indentation since the previous `if` statement returns from the function

mii/client.py
```diff
@@ -156,6 +166,48 @@ def destroy_session(self, session_id):
...
+    def __init__(self, task, deployment_name):
+        self.task = task
+        self.deployment_name = deployment_name
+
+    def query(self, request_dict, **query_kwargs):
+        task_methods = GRPC_METHOD_TABLE[self.task]
+        inference_pipeline = mii.non_persistent_models[self.deployment_name][0]
+
+        if self.task == Tasks.QUESTION_ANSWERING:
+            if 'question' not in request_dict or 'context' not in request_dict:
```
mrwyattii, line 179: I think we should error out here. If the user does not provide the correct keywords, we should not fail silently

mii/client.py
```diff
@@ -156,6 +166,48 @@ def destroy_session(self, session_id):
...
+                return
+
+            return task_methods.run_inference(inference_pipeline,
+                                              request_dict,
+                                              query_kwargs,
+                                              True)
```
mrwyattii, line 185: What is this `True`? This should cause an error since `run_inference` only takes 3 arguments.

mii/client.py
```diff
@@ -156,6 +166,48 @@ def destroy_session(self, session_id):
...
+            text = request_dict['text']
+            conversation_id = request_dict['conversation_id']
+            past_user_inputs = request_dict['past_user_inputs']
+            generated_responses = request_dict['generated_responses']
+            conv = Conversation(text=text,
+                                conversation_id=conversation_id,
+                                past_user_inputs=past_user_inputs,
+                                generated_responses=generated_responses,
+                                **kwargs)
+            args = (conv, )
```
mrwyattii, line 198: I wonder if we could refactor the Conversation task methods so we could re-use the code here:
https://github.com/microsoft/DeepSpeed-MII/blob/dc5ab44dfa48ae9f0a99b356e96c8849c0c78aea/mii/method_table.py#L192

Maybe bring the `conv = Conversation(...)` bit out into it's own function called `create_conversation` and use `getattr(request, "conversation_id"), getattr(request, "past_user_inputs"), ...` so that the code works with the proto message and the query dictionary.

mii/method_table.py
```diff
@@ -54,8 +54,11 @@ def pack_request_to_proto(self, request_dict, **query_kwargs):
     def unpack_request_from_proto(self, request):
         return request
 
-    def run_inference(self, inference_pipeline, args, kwargs):
-        return inference_pipeline(*args, **kwargs)
+    def run_inference(self, inference_pipeline, args, kwargs, is_non_persistent=False):
```
mrwyattii, line 57: We should be able to avoid needing this extra `is_non_persistent` if we modify the `args` value in `MIINonPersistentClient`

mrwyattii (changes requested):

mii/client.py
TosinSeg: There is an error for some of the non_persistent tests when I try and pass *args, it will take the string that is passed in, and treat every character in the string as an argument and throw an error.

mii/method_table.py
TosinSeg: I am currently modifying the args value like this in the MIINonPersistentClient class. 
```python 
else:
            query = request_dict['query']
            kwargs = {}
            return task_methods.run_inference(inference_pipeline, query, kwargs, True)
```
But even with this, there is an error when I try to pass query in as *args to the run_inference call in method_table (for the non_persistent models).

mii/client.py
```diff
@@ -34,6 +34,11 @@ def mii_query_handle(deployment_name):
     Returns:
         query_handle: A query handle with a single method `.query(request_dictionary)` using which queries can be sent to the model.
     """
+    if deployment_name in mii.non_persistent_models:
+        assert deployment_name in mii.non_persistent_models, f"Could not find '{deployment_name}'"
```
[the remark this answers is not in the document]
mrwyattii: Don't need this assert since this is already checked in the line above

mii/client.py
```diff
@@ -156,6 +161,38 @@ def destroy_session(self, session_id):
...
+            if 'question' not in request_dict or 'context' not in request_dict:
+                raise Exception(
+                    "Question Answering Task requires 'question' and 'context' keys")
+            args = (request_dict["question"], request_dict["context"])
+            kwargs = query_kwargs
+
+        elif self.task == Tasks.CONVERSATIONAL:
+            conv = create_conversation(request_dict, **query_kwargs)
+            args = (conv, )
+            kwargs = query_kwargs
```
mrwyattii, line 183:
```suggestion
            kwargs = {}
```

take a look here:
https://github.com/microsoft/DeepSpeed-MII/blob/dc5ab44dfa48ae9f0a99b356e96c8849c0c78aea/mii/method_table.py#L198

We want `kwargs` to be empty since we're using them above in `create_conversation`

mii/utils.py
```diff
@@ -209,6 +209,41 @@ def get_num_gpus(mii_configs):
...
+        conversation_id = getattr(request, 'conversation_id')
+        past_user_inputs = getattr(request, 'past_user_inputs')
+        generated_responses = getattr(request, 'generated_responses')
+
+    conv = Conversation(text=text,
+                        conversation_id=conversation_id,
+                        past_user_inputs=past_user_inputs,
+                        generated_responses=generated_responses,
+                        **kwargs)
+    return conv
```
mrwyattii, line 244: I think this should be part of the `ConversationMethods` class rather than here in the utils.

tests/test_non_persistent_deployment.py
```diff
@@ -0,0 +1,231 @@
...
+        with open(hostfile, "w") as f:
+            f.write(f"localhost slots={num_gpu}")
+
+    return {
+        'dtype': dtype,
+        'tensor_parallel': tensor_parallel,
+        'load_with_sys_mem': load_with_sys_mem,
+        'enable_load_balancing': enable_load_balancing,
+        'replica_num': num_gpu * enable_load_balancing // tensor_parallel,
+        'hostfile': hostfile,
```
mrwyattii, line 37: I think we can leave these out since load balancing isn't supported (but keep `enable_load_balancing` to ensure we get the expected error)

tests/test_non_persistent_deployment.py
```diff
@@ -0,0 +1,231 @@
...
+            "distilgpt2",
+            {
+                "query": "DeepSpeed is the greatest"
+            },
+        ),
+    ],
+)
+def test_zero_config_fail(non_persistent_deployment, query):
+    print(non_persistent_deployment)
+    assert "MII Config Error" in str(non_persistent_deployment.value)
```
mrwyattii, line 231: We can leave these out. We don't want to unnecessarily increase the time of our unit tests and the code paths tested here will be covered in `test_local_deployment.py`

tests/test_non_persistent_deployment.py
```diff
@@ -0,0 +1,231 @@
...
+    hostfile = os.path.join(tmpdir, "hostfile")
+    num_gpu = torch.cuda.device_count()
+    enable_load_balancing = enable_load_balancing
+    if enable_load_balancing:
+        with open(hostfile, "w") as f:
+            f.write(f"localhost slots={num_gpu}")
+
+    return {
+        'dtype': dtype,
+        'tensor_parallel': tensor_parallel,
```
mrwyattii, line 33: Can we run with tensor parallelism with this deployment type? I think we will have to run with `deepspeed --num_gpus X mii_example.py` for tensor parallelism to work with this new deployment type. We should add a check in `deployment.py` that asserts that `int(os.genenv["WORLD_SIZE"]) == mii_config.tensor_parallel` for the non-persistent deployment only.

mrwyattii (changes requested):

mii/method_table.py
```diff
@@ -95,7 +94,7 @@ def preprocess_session(self, session_id, args):
         args = ([self.session_context[session_id] + args[0][0]], )
         return args
 
-    def run_inference(self, inference_pipeline, args, kwargs):
+    def run_inference(self, inference_pipeline, args, kwargs, is_non_persistent=False):
```
mrwyattii, line 97:
```suggestion
    def run_inference(self, inference_pipeline, args, kwargs):
```

tests/test_non_persistent_deployment.py
```diff
@@ -0,0 +1,159 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+import pytest
+import os
+import torch
+from types import SimpleNamespace
+from .utils import *  # noqa: F401
+import mii
+''' These fixtures provide a local deployment and ensure teardown '''
```
mrwyattii, line 11:
```suggestion
```

mrwyattii (approved):

mii/client.py
```diff
@@ -34,6 +34,10 @@ def mii_query_handle(deployment_name):
     Returns:
         query_handle: A query handle with a single method `.query(request_dictionary)` using which queries can be sent to the model.
     """
+    if deployment_name in mii.non_persistent_models:
+        inference_pipeline, task = mii.non_persistent_models[deployment_name]
```
tohtana, line 38: Shouldn't we explicitly specify if we want to use a non-persistent model or not?

mii/client.py
```diff
@@ -156,6 +160,38 @@ def destroy_session(self, session_id):
...
 
 
+class MIINonPersistentClient():
+    def __init__(self, task, deployment_name):
+        self.task = task
+        self.deployment_name = deployment_name
+
+    def query(self, request_dict, **query_kwargs):
+        task_methods = GRPC_METHOD_TABLE[self.task]
+        inference_pipeline = mii.non_persistent_models[self.deployment_name][0]
```
tohtana, line 170: We should throw an error that explains the reason when the `deployment_name` is not in the dictionary.

The another possible design is to return a client object from `deploy()` (or a new function like `deploy_non_persistent()`. The user wouldn't need to be aware of the deployment name, which is globally and internally stored in MII.

mii/client.py
```diff
@@ -156,6 +160,38 @@ def destroy_session(self, session_id):
...
+
+        else:
+            args = (request_dict['query'], )
+            kwargs = query_kwargs
+
+        return task_methods.run_inference(inference_pipeline, args, query_kwargs)
+
+    def terminate(self):
+        print("Terminating ...")
+        del mii.non_persistent_models[self.deployment_name]
```
tohtana, line 192: It would be helpful if we contain the deployment name in the message.

tohtana:
Shouldn't we add an example in `examples/local` and some description to README (e.g. following "RESTful API support")?

mii/client.py
TosinSeg: I will add this in now.

mii/client.py
TosinSeg: I added an assertion for this in the query_handle() function.

mrwyattii: We still need to add an assertion here that `self.deployment_name in miii.non_persistent_models` because a user could still call `terminate` and then try run a query after, resulting in a potentially confusing error.

mii/deployment.py
```diff
@@ -160,6 +175,29 @@ def _deploy_aml(deployment_name, model_name, version):
...
+    deploy(task,
+           model,
+           deployment_name,
+           deployment_type,
+           model_path,
+           enable_deepspeed,
+           enable_zero,
+           ds_config,
+           mii_config,
+           version)
```
mrwyattii, line 197:
```suggestion
    deploy(task,
           model,
           deployment_name,
           **kwargs)
```

mii/deployment.py
```diff
@@ -160,6 +175,29 @@ def _deploy_aml(deployment_name, model_name, version):
...
+def deploy_non_persistent(task,
+                          model,
+                          deployment_name,
+                          deployment_type=DeploymentType.LOCAL,
+                          model_path=None,
+                          enable_deepspeed=True,
+                          enable_zero=False,
+                          ds_config=None,
+                          mii_config={},
+                          version=1):
```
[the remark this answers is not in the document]
tohtana: If we introduce "deploy_non_persistent" which returns a client object, can we have an inference pipeline in the object instead of having it globally? We could tie the lifetimes of the client, an inference pipeline, and the deployment name.

mii/deployment.py
```diff
@@ -125,21 +126,35 @@ def deploy(task,
...
+                          model_path=model_path,
+                          lb_config=lb_config)
 
     if deployment_type == DeploymentType.AML:
         _deploy_aml(deployment_name=deployment_name, model_name=model, version=version)
     elif deployment_type == DeploymentType.LOCAL:
         return _deploy_local(deployment_name, model_path=model_path)
+    elif deployment_type == DeploymentType.NON_PERSISTENT:
+        assert not mii_config.enable_load_balancing, "Cannot use Load Balancing with Non persistent deployment"
+        assert int(os.getenv('WORLD_SIZE', '1')) == mii_config.tensor_parallel, "World Size does not equal number of tensors"
```
mrwyattii, line 147:
```suggestion
        assert int(os.getenv('WORLD_SIZE', '1')) == mii_config.tensor_parallel, "World Size does not equal number of tensors. When using non-persistent deployment type, please launch with `deepspeed --num_gpus <tensor_parallel>`"
```

mii/client.py
TosinSeg: Added it now, and removed the extra parameter in the query_handle().

README.md
```diff
@@ -178,6 +178,24 @@ mii.deploy(...
...
+
+```python
+...
+mii.deploy(deployment_name = DEPLOYMENT_NAME,
+	   deployment_type=mii.constants.DeploymentType.NON_PERSISTENT
+	   ...
+	   )
+
+generator = mii.mii_query_handle(DEPLOYMENT_NAME)
+retult = generator.query({"query": ["DeepSpeed is", "Seattle is"]}, do_sample=True, max_new_tokens=30})
```
mrwyattii, line 193:
```suggestion
result = generator.query({"query": ["DeepSpeed is", "Seattle is"]}, do_sample=True, max_new_tokens=30})
```

README.md
```diff
@@ -178,6 +178,24 @@ mii.deploy(...
     mii_config=mii_configs)
 ```
 
+**Non-persistent Deployment**
+
+You can enable a non-persistent deployment which allows you to make queries without standing up a server. This can be achieved by setting `deployment_type` to `NON_PERSISTENT` in `mii.deploy(...)`.
```
mrwyattii, line 183:
```suggestion
You can enable a non-persistent deployment which allows you to make queries without standing up a server. The non-persistent deployment acts as a simplified interface to DeepSpeed-inference for use cases that do not require creating a persistent model server process. Changing the `deployment_type` to `NON_PERSISTENT` in `mii.deploy(...)` will activate this option.
```

mrwyattii (approved):

PR #200, opened 2023-06-02
Refactor task methods

mrwyattii:
We're aiming to add a non-GRPC deployment method in #197. In order to facilitate that, refactoring the different task methods so we can reuse code when creating conversation sessions, processing kwargs, and calling the inference pipeline.


mrwyattii:
Replaced by #202, closing

PR #205, opened 2023-06-20
Always enable load balancing

TosinSeg:
Always enable load balancing


mii/client.py
```diff
@@ -40,13 +40,14 @@ def mii_query_handle(deployment_name):
         return MIINonPersistentClient(task, deployment_name)
 
     task_name, mii_configs = _get_deployment_info(deployment_name)
-    if mii_configs.enable_load_balancing:
-        return MIIClient(task_name, "localhost", mii_configs.port_number)
+    return MIIClient(task_name, "localhost", mii_configs.port_number)
+    """
```
mrwyattii, line 44: Let's remove this commented code

mii/server.py
```diff
@@ -48,7 +47,7 @@ def __init__(self,
 
         self.port_number = mii_configs.port_number
 
-        if mii_configs.enable_load_balancing and mii_configs.hostfile is None:
+        if mii_configs.hostfile is None:
             raise ValueError(
                 "hostfile must be provided if enable_load_balancing == True")
```
mrwyattii, line 52: This error message will not make sense now

tests/test_local_deployment.py
```diff
@@ -38,26 +38,22 @@ def mii_configs(
...
+        f.write(f"localhost slots={num_gpu}")
 
     return {
         'dtype': dtype,
         'tensor_parallel': tensor_parallel,
         'port_number': port_number,
         'load_with_sys_mem': load_with_sys_mem,
-        'enable_load_balancing': enable_load_balancing,
-        'replica_num': num_gpu * enable_load_balancing // tensor_parallel,
+        'replica_num': num_gpu * 1 // tensor_parallel,
```
mrwyattii, line 56:
```suggestion
        'replica_num': num_gpu // tensor_parallel,
```

mrwyattii (changes requested):
I think we should also remove the `enable_load_balancing` pytest fixture.

Additionally, we should ensure that MII will work correctly if a hostfile is not specified. In the tests, we generate the hostfile, but perhaps this sort of functionality should be embedded in MII so that users are not required to generate one (especially if they are using just a single replica)

mii/deployment.py
```diff
@@ -106,26 +106,25 @@ def deploy(task,
 
     # add fields for replica deployment
     lb_config = None
```
tohtana, line 108: Remove this line. `lb_config` is always initialized later.

mii/server.py
```diff
@@ -60,11 +63,7 @@ def __init__(self,
                                              ds_config,
                                              mii_configs,
                                              lb_config)
-        deployment = lb_config.replica_configs if mii_configs.enable_load_balancing else [
-            ReplicaConfig(hostname='localhost',
-                          tensor_parallel_ports=[mii_configs.port_number],
-                          torch_dist_port=mii_configs.torch_dist_port)
-        ]
+        deployment = lb_config.replica_configs
```
tohtana, line 66: Now we don't need `deployment`

mrwyattii (approved):

mrwyattii (approved):

tohtana (approved):

PR #208, opened 2023-06-27
Multi model deployment

TosinSeg:


mii/deployment.py
```diff
@@ -66,37 +67,49 @@ def deploy(task,
         If deployment_type is `LOCAL`, returns just the name of the deployment that can be used to create a query handle using `mii.mii_query_handle(deployment_name)`
 
     """
+    if len(deployments == 0):
+        assert model is not None and task is not None and deployment_name is not None, "model, task, and deployment name must be set to deploy sigular model"
```
mrwyattii, line 71:
```suggestion
    if not deployments:
        assert all((task, model, deployment_name)), "model, task, and deployment name must be set to deploy sigular model"
```

mii/config.py
```diff
@@ -123,4 +124,16 @@ class LoadBalancerConfig(BaseModel):
...
+
+
+class Deployment(BaseModel):
+    deployment_name: str
+    task: str
+    model: str
+    enable_deepspeed: bool = True
+    enable_zero: bool = True
+    GPU_index_map: dict = None
+    mii_config: dict = None
```
mrwyattii, line 137: Pydantic models can be nested, so you can change this type to `MIIConfig`.

mii/models/score/score_template.py
```diff
@@ -16,6 +16,8 @@
 
 def init():
     model_path = mii.utils.full_model_path(configs[mii.constants.MODEL_PATH_KEY])
+    deployment_tag = configs[mii.constants.DEPLOYMENT_TAG_KEY]
+    deployments = mii.multi_model_deployments[deployment_tag]
```
mrwyattii, line 20: I don't think this will work. I believe the `mii.multi_model_deployments` that you add to in `deployment.py` is on a different process than the one that will be running the generated `score.py` file. So this will just be an empty dict when the score file gets executed!

TosinSeg: Oh, ok that makes sense!

mii/grpc_related/modelresponse_server.py
```diff
@@ -190,41 +200,78 @@ def choose_stub(self, call_count):
     def intercept_service(self, continuation, handler_call_details):
         next_handler = continuation(handler_call_details)
         assert next_handler.unary_unary is not None
+        #USE KWARGS LIKE THEY ARE USED TO MAKE SESSIONS TO GET THE DEPLOYMENT NAME TO HASH THE COUNTERS/STUBS
 
         def invoke_intercept_method(request_proto, context):
             method_name = _get_grpc_method_name(handler_call_details.method)
+            kwargs = unpack_proto_query_kwargs(request_proto.query_kwargs)
```
mrwyattii, line 207: I don't think this is required. I believe we can grab the deployment name from kwargs here: https://github.com/microsoft/DeepSpeed-MII/blob/ce80314c0d086485b1559663102c48d2c24cb2eb/mii/client.py#L125

The `query_kwargs` is intended to be passed to the pipeline object - so using it to also capture the deployment name breaks that assumption.

mii/client.py
```diff
@@ -55,24 +67,36 @@ class MIIClient():
...
+        task_methods = GRPC_METHOD_TABLE[task]
         proto_request = task_methods.pack_request_to_proto(request_dict, **query_kwargs)
         proto_response = await getattr(self.stub, task_methods.method)(proto_request)
         return task_methods.unpack_response_from_proto(proto_response)
 
-    def query(self, request_dict, **query_kwargs):
+    def query(self, request_dict, deployment_name=None, **query_kwargs):
+        task = None
+        if deployment_name is None:  #mii.terminate() or single model
+            deployment_name = self.deployments[0][mii.constants.DEPLOYMENT_NAME_KEY]
```
TosinSeg, line 89: Check for multiple models is there is not deployment name passed in.

mii/terminate.py
```diff
@@ -7,21 +7,21 @@
...
-    if (deployment_name in mii.non_persistent_models):
+def terminate(deployment_tag):
+    mii.utils.logger.info(f"Terminating server for {deployment_tag}")
+    generator = mii.mii_query_handle(deployment_tag)
+    if (deployment_tag in mii.non_persistent_models):
         generator.terminate()
         return
     try:
-        generator.query({'query': ''})
+        generator.query({'query': ''}, None)
```
TosinSeg, line 17: Use a special string that signifies that we are running terminate.

mii/client.py
```diff
@@ -12,17 +12,27 @@
...
+    deployments = []
+    configs = mii.utils.import_score_file(deployment_tag).configs
+    for deployment in configs:
+        if not isinstance(configs[deployment], dict):
+            continue
+        deployments.append(configs[deployment])
+        mii_configs_dict = configs[deployment][mii.constants.MII_CONFIGS_KEY]
+        mii_configs = mii.config.MIIConfig(**mii_configs_dict)
+    return deployments
+    """
```
mrwyattii, line 25: Remove commented code

mii/client.py
```diff
@@ -12,17 +12,27 @@
...
-    mii_configs_dict = configs[mii.constants.MII_CONFIGS_KEY]
+def _get_deployment_info(deployment_tag):
+    deployments = []
+    configs = mii.utils.import_score_file(deployment_tag).configs
+    for deployment in configs:
+        if not isinstance(configs[deployment], dict):
+            continue
+        deployments.append(configs[deployment])
+        mii_configs_dict = configs[deployment][mii.constants.MII_CONFIGS_KEY]
+        mii_configs = mii.config.MIIConfig(**mii_configs_dict)
```
mrwyattii, line 23: These are unused, why do we do this?

mii/client.py
```diff
@@ -12,17 +12,27 @@
...
 
-def _get_deployment_info(deployment_name):
-    configs = mii.utils.import_score_file(deployment_name).configs
-    task = configs[mii.constants.TASK_NAME_KEY]
-    mii_configs_dict = configs[mii.constants.MII_CONFIGS_KEY]
+def _get_deployment_info(deployment_tag):
+    deployments = []
+    configs = mii.utils.import_score_file(deployment_tag).configs
+    for deployment in configs:
+        if not isinstance(configs[deployment], dict):
```
mrwyattii, line 19: Is there a reason that these would not be a `dict`?

mii/client.py
```diff
@@ -12,17 +12,27 @@
 from mii.method_table import GRPC_METHOD_TABLE
 
 
-def _get_deployment_info(deployment_name):
-    configs = mii.utils.import_score_file(deployment_name).configs
-    task = configs[mii.constants.TASK_NAME_KEY]
-    mii_configs_dict = configs[mii.constants.MII_CONFIGS_KEY]
+def _get_deployment_info(deployment_tag):
+    deployments = []
```
mrwyattii, line 16: We could rename this to `deployment_configs` to be more accurate. We don't want to get confused about what this list holds.

mrwyattii: This change could probably be propagated to the rest of the code too (unless I'm misunderstanding what is held in this variable)

mii/client.py
```diff
@@ -55,24 +67,37 @@ class MIIClient():
...
         proto_response = await getattr(self.stub, task_methods.method)(proto_request)
         return task_methods.unpack_response_from_proto(proto_response)
 
-    def query(self, request_dict, **query_kwargs):
+    def query(self, request_dict, deployment_name=None, **query_kwargs):
+        task = None
+        if deployment_name is None:  #mii.terminate() or single model
+            #assert len(self.deployments) == 1, "Must pass deployment_name to query when using multiple deployments"
+            deployment_name = self.deployments[0]['deployment_name']
+            task = get_task(self.deployments[0]['task_name'])
```
mrwyattii, line 91: We should require that users provide `deployment_name` unless there is only a single depoyment

mii/client.py
```diff
@@ -55,24 +67,37 @@ class MIIClient():
...
+        else:
+            for deployment in self.deployments:
+                if deployment[mii.constants.DEPLOYMENT_NAME_KEY] == deployment_name:
+                    task = get_task(deployment[mii.constants.TASK_NAME_KEY])
+                    break
```
mrwyattii, line 96:
```suggestion
            for deployment in self.deployments:
                if deployment[mii.constants.DEPLOYMENT_NAME_KEY] == deployment_name:
                    task = get_task(deployment[mii.constants.TASK_NAME_KEY])
                    break
            else:
                assert False, f"{deployment_name} not found in list of deployments"
```

If a user provides a deployment name that does not exist, we need to error out here

mii/client.py
```diff
@@ -55,24 +67,37 @@ class MIIClient():
...
+        query_kwargs['deployment_name'] = deployment_name
```
mrwyattii, line 97: `query_kwargs` is intended to contain all the user-passed kwargs that should be passed to the inference pipeline. I don't think we should store the deployment name in this dictionary

mii/client.py
```diff
@@ -86,17 +111,35 @@ async def create_session_async(self, session_id):
...
+    def create_session(self, session_id, deployment_name=None):
+        task = None
+        if deployment_name is None:  #mii.terminate() or single model
+            deployment_name = self.deployments[0][mii.constants.DEPLOYMENT_NAME_KEY]
+            task = get_task(self.deployments[0][mii.constants.TASK_NAME_KEY])
+        else:
+            for deployment in self.deployments:
+                if deployment[mii.constants.DEPLOYMENT_NAME_KEY] == deployment_name:
+                    task = get_task(deployment[mii.constants.TASK_NAME_KEY])
+                    break
```
mrwyattii, line 123: This looks like a copy of the code above in `query()` - we should make it a class method to avoid copied code

mii/client.py
```diff
@@ -86,17 +111,35 @@ async def create_session_async(self, session_id):
...
+    def destroy_session(self, session_id, deployment_name=None):
+        task = None
+        if deployment_name is None:  #mii.terminate() or single model
+            deployment_name = self.deployments[0][mii.constants.DEPLOYMENT_NAME_KEY]
+            task = get_task(self.deployments[0][mii.constants.TASK_NAME_KEY])
+        else:
+            for deployment in self.deployments:
+                if deployment[mii.constants.DEPLOYMENT_NAME_KEY] == deployment_name:
+                    task = get_task(deployment[mii.constants.TASK_NAME_KEY])
+                    break
```
mrwyattii, line 141: Same comment as above, let's avoid copied code blocks

mii/client.py
```diff
@@ -188,7 +231,10 @@ def terminate(self):
         del mii.non_persistent_models[self.deployment_name]
 
 
-def terminate_restful_gateway(deployment_name):
-    _, mii_configs = _get_deployment_info(deployment_name)
-    if mii_configs.enable_restful_api:
-        requests.get(f"http://localhost:{mii_configs.restful_api_port}/terminate")
+def terminate_restful_gateway(deployment_tag):
+    deployments = _get_deployment_info(deployment_tag)
+    for deployment in deployments:
```
mrwyattii, line 236: Do we support having separate restful APIs for each deployment in a multi-model deployment? I think it is tied to the load balancer, not the inference servers.

mii/config.py
```diff
@@ -123,4 +124,18 @@ class LoadBalancerConfig(BaseModel):
 
     class Config:
         validate_all = True
-        validate_assignment = True
+
+
+validate_assignment = True
+
+
+class Deployment(BaseModel):
```
mrwyattii, line 132: Rename this to `DeploymentConfig`

mii/deployment.py
```diff
@@ -105,47 +127,55 @@ def deploy(task,
...
+        replica_pool = _allocate_processes(mii_config.hostfile,
+                                           mii_config.tensor_parallel,
+                                           mii_config.replica_num,
+                                           deployment.GPU_index_map)
+
+        for i, (hostname, gpu_indices) in enumerate(replica_pool):
+            # Reserver port for a LB proxy when replication is enabled
+            if hostname not in port_map:
+                port_map[hostname] = set()
+            port_offset = 1
```
mrwyattii, line 143: Move this line outside the loop if it stays constant

mii/grpc_related/modelresponse_server.py
```diff
@@ -191,40 +199,82 @@ def intercept_service(self, continuation, handler_call_details):
...
+            if method_name != TERMINATE_METHOD:
+                assert "deployment_name" in kwargs, "Must include deployment_name in kwargs for query"
+            deployment_name = kwargs.get('deployment_name')
+            kwargs.pop('deployment_name', None)
+            task = None
+            for repl in self.replica_configs:
+                if repl.deployment_name == deployment_name:
+                    task = repl.task
+                    break
```
mrwyattii, line 222: This should probably be in `__init__`. We store a dictionary mapping `deployment_name` to `task` and avoid doing this computation every query

mii/grpc_related/modelresponse_server.py
```diff
@@ -191,40 +199,82 @@ def intercept_service(self, continuation, handler_call_details):
...
+                request_dict['question'] = str(request_proto.question)
+                request_dict['context'] = str(request_proto.context)
+                new_request = method.pack_request_to_proto(request_dict, **kwargs)
+            else:
+                request_dict = {}
+                request_dict["query"] = list(
+                    request_proto.request
+                ) if method_name == "GeneratorReply" or method_name == "Txt2ImgReply" else str(
+                    request_proto.request)
+                new_request = method.pack_request_to_proto(request_dict, **kwargs)
```
mrwyattii, line 246: This code looks like it is copied from `methods_table.py`:
https://github.com/microsoft/DeepSpeed-MII/blob/ce80314c0d086485b1559663102c48d2c24cb2eb/mii/method_table.py#L218

Can we find a way to reuse the code blocks we already have that handle this work?

mii/grpc_related/modelresponse_server.py
```diff
@@ -191,40 +199,82 @@ def intercept_service(self, continuation, handler_call_details):
...
 
-            kwargs = unpack_proto_query_kwargs(request_proto.query_kwargs)
             if "session_id" in kwargs:
                 session_id = kwargs["session_id"]
                 if session_id not in self.replica_sessions:
                     raise ValueError(f"session not found")
                 replica_index = self.replica_sessions[session_id]
 
-            ret = self.stubs[replica_index].invoke(method_name, request_proto)
+            assert new_request is not None, "test"
```
mrwyattii, line 274: Is this a debug assert?

mii/models/score/generate.py
```diff
@@ -9,33 +9,39 @@
...
         config_dict[mii.constants.LOAD_BALANCER_CONFIG_KEY] = lb_config
 
-    if len(mii.__path__) > 1:
-        logger.warning(
-            f"Detected mii path as multiple sources: {mii.__path__}, might cause unknown behavior"
-        )
+        if len(mii.__path__) > 1:
+            logger.warning(
+                f"Detected mii path as multiple sources: {mii.__path__}, might cause unknown behavior"
+            )
```
mrwyattii, line 44: Was this accidently indented?

mii/models/score/generate.py
```diff
@@ -9,33 +9,39 @@
...
+        config_dict[deployment.deployment_name][
+            mii.constants.MODEL_NAME_KEY] = deployment.model
+        config_dict[deployment.deployment_name][
+            mii.constants.ENABLE_DEEPSPEED_KEY] = deployment.enable_deepspeed
+        config_dict[deployment.deployment_name][
+            mii.constants.MII_CONFIGS_KEY] = deployment.mii_config.dict()
+        config_dict[deployment.deployment_name][
+            mii.constants.ENABLE_DEEPSPEED_ZERO_KEY] = deployment.enable_zero
+        config_dict[deployment.deployment_name][
+            mii.constants.DEEPSPEED_CONFIG_KEY] = deployment.ds_config
```
mrwyattii, line 36:
```suggestion
        deployment_config = {
            mii.constants.DEPLOYMENT_NAME_KEY: deployment.deployment_name,
            mii.constants.TASK_NAME_KEY: mii.utils.get_task_name(deployment.task),
            mii.constants.MODEL_NAME_KEY: deployment.model,
            mii.constants.ENABLE_DEEPSPEED_KEY: deployment.enable_deepspeed,
            mii.constants.MII_CONFIGS_KEY: deployment.mii_config.dict(),
            mii.constants.ENABLE_DEEPSPEED_ZERO_KEY: deployment.enable_zero,
            mii.constants.DEEPSPEED_CONFIG_KEY: deployment.ds_config,
        }
        config_dict[deployment.deployment_name] = deployment_config
```

We can make this cleaner and easier to read

mrwyattii (changes requested):

examples/multi_model/query.py
```diff
@@ -0,0 +1,46 @@
...
+
+result = generator.query({
+    'question':
+    "What is the greatest?",
+    'context':
+    "DeepSpeed is the greatest",
+    "deployment_name":
+    "deepset/roberta-large-squad2" + "-qa-deployment"
+})
+results.append(result)
```
mrwyattii, line 46: Add print statements to show the results

mii/client.py
```diff
@@ -8,21 +8,38 @@
...
-def mii_query_handle(deployment_name):
+from mii.deployment import allocate_processes, create_score_file
+from mii.config import DeploymentConfig
+
+
+def _get_deployment_configs(deployment_tag):
+    deployments = {}
+    configs = mii.utils.import_score_file(deployment_tag).configs
+    for deployment in configs[mii.constants.DEPLOYMENTS_KEY].values():
+        deployment[mii.constants.DEPLOYED_KEY] = True
```
mrwyattii, line 21: What is the need for this hardcoded `True`? This seems like something we should be able to remove

mii/client.py
```diff
@@ -8,21 +8,38 @@
...
+            'model': deployment[mii.constants.MODEL_NAME_KEY],
+            'enable_deepspeed': deployment[mii.constants.ENABLE_DEEPSPEED_KEY],
+            'enable_zero': deployment[mii.constants.ENABLE_DEEPSPEED_ZERO_KEY],
+            'GPU_index_map': None,
+            'mii_config': deployment[mii.constants.MII_CONFIGS_KEY],
+            'ds_config': deployment[mii.constants.DEEPSPEED_CONFIG_KEY],
+            'version': 1,
+            'deployed': deployment[mii.constants.DEPLOYED_KEY]
+        }
+        deployments[deployment_name] = DeploymentConfig.parse_obj(data)
```
mrwyattii, line 35:
```suggestion
        deployments[deployment_name] = DeploymentConfig(**data)
```

mii/client.py
```diff
@@ -8,21 +8,38 @@
...
+    for deployment in configs[mii.constants.DEPLOYMENTS_KEY].values():
+        deployment[mii.constants.DEPLOYED_KEY] = True
+        deployment_name = deployment[mii.constants.DEPLOYMENT_NAME_KEY]
+        data = {
+            'deployment_name': deployment[mii.constants.DEPLOYMENT_NAME_KEY],
+            'task': deployment[mii.constants.TASK_NAME_KEY],
+            'model': deployment[mii.constants.MODEL_NAME_KEY],
+            'enable_deepspeed': deployment[mii.constants.ENABLE_DEEPSPEED_KEY],
+            'enable_zero': deployment[mii.constants.ENABLE_DEEPSPEED_ZERO_KEY],
+            'GPU_index_map': None,
```
mrwyattii, line 29: Do we throw away the GPU map?

mii/client.py
```diff
@@ -35,12 +52,23 @@ def mii_query_handle(deployment_name):
...
-        inference_pipeline, task = mii.non_persistent_models[deployment_name]
-        return MIINonPersistentClient(task, deployment_name)
+    if deployment_tag in mii.non_persistent_models:
+        inference_pipeline, task = mii.non_persistent_models[deployment_tag]
+        return MIINonPersistentClient(task, deployment_tag)
+
+    deployments, lb_config, model_path, port_map = _get_deployment_configs(deployment_tag)
+    mii_configs = None
+    if len(deployments) > 0:
+        mii_configs = next(iter(deployments.values())).mii_config
```
mrwyattii, line 62: Does this assume that all the `mii_config` have the same port number value? Perhaps we need to rethink what the port number field in the `MIIConfig` is meant for with support for multi-model deployments

mii/client.py
```diff
@@ -55,24 +83,59 @@ class MIIClient():
...
-    async def _request_async_response(self, request_dict, **query_kwargs):
-        if self.task not in GRPC_METHOD_TABLE:
-            raise ValueError(f"unknown task: {self.task}")
-
-        task_methods = GRPC_METHOD_TABLE[self.task]
+        self.stub = None
+        self.host = host
+        if port is not None:
+            channel = create_channel(host, port)
+            self.stub = modelresponse_pb2_grpc.DeploymentManagementStub(channel)
```
mrwyattii, line 99: Is this to support lazy loading deployments (or deployments with no models)?

mii/deployment.py
```diff
@@ -66,86 +68,98 @@ def deploy(task,
...
 
+    if not deployments and not all((model, task, deployment_name)):
+        assert deployment_tag is not None, "Deployment tag must be set when starting empty deployment"
+        create_score_file(deployment_tag=deployment_tag,
+                          deployment_type=deployment_type,
+                          deployments=None,
+                          model_path=model_path,
+                          port_map=None,
+                          lb_config=None)
+        return None
```
mrwyattii, line 84: This is a messy way to add this logic into the deployment function. We can refactor this as such:
```python
if deployment and any((model, task, deployment_name)):
    # Error out, it is ambiguous what we should do here
if all((model, task, deployment_name)):
    deployments.append(DeploymentConfig(...))
    if not deployment_tag:
        deployment_tag = deployment_name
assert deployment_tag, "deployment_tag must be set for multi-model"
for deployment in deployments:
    ...
...
create_score_file(...)
```

mii/deployment.py
```diff
@@ -66,86 +68,98 @@ def deploy(task,
...
+        mii_config = deployment.mii_config
+        if deployment.enable_zero:
+            if deployment.ds_config.get("fp16", {}).get("enabled", False):
+                assert (mii_config.dtype == torch.half), "MII Config Error: MII dtype and ZeRO dtype must match"
+            else:
+                assert (mii_config.dtype == torch.float), "MII Config Error: MII dtype and ZeRO dtype must match"
+        assert not (enable_deepspeed and enable_zero), "MII Config Error: DeepSpeed and ZeRO cannot both be enabled, select only one"
 
     # aml only allows certain characters for deployment names
     if deployment_type == DeploymentType.AML:
```
mrwyattii, line 114: Let's add an assert that prevents multi-model for AML deployments for the time being. We don't have a great way of testing this.

mii/grpc_related/modelresponse_server.py
```diff
@@ -34,6 +33,16 @@ def get_stop_event(self):
         return self._stop_event
 
 
+class DeploymentManagement(ServiceBase,
+                           modelresponse_pb2_grpc.DeploymentManagementServicer):
+    def AddDeployment(self, request, context):
+        print("DEPLOYMENT ADDED")
```
mrwyattii, line 39: Remove debug statement

mii/grpc_related/modelresponse_server.py
```diff
@@ -191,40 +207,79 @@ def intercept_service(self, continuation, handler_call_details):
         next_handler = continuation(handler_call_details)
         assert next_handler.unary_unary is not None
 
+        #USE KWARGS LIKE THEY ARE USED TO MAKE SESSIONS TO GET THE DEPLOYMENT NAME TO HASH THE COUNTERS/STUBS
+
```
mrwyattii, line 211:
```suggestion
```

mii/grpc_related/modelresponse_server.py
```diff
@@ -191,40 +207,79 @@ def intercept_service(self, continuation, handler_call_details):
...
         def invoke_intercept_method(request_proto, context):
             method_name = _get_grpc_method_name(handler_call_details.method)
+            if method_name == ADD_DEPLOYMENT_METHOD:
+                deployment_name = str(getattr(request_proto, "deployment_name"))
+                if deployment_name not in self.stubs:
```
mrwyattii, line 216: What happens if the deployment name is in `self.stubs`? Should we error out?

mii/grpc_related/proto/modelresponse.proto
```diff
@@ -36,6 +36,21 @@ service ModelResponse {
...
+  rpc AddDeployment(AddDeployRequest) returns (google.protobuf.Empty) {}
+  rpc CreateSession (SessionID) returns (google.protobuf.Empty) {}
+  rpc DestroySession (SessionID) returns (google.protobuf.Empty) {}
+  rpc GeneratorReply (MultiStringRequest) returns (MultiStringReply) {}
+  rpc ClassificationReply (SingleStringRequest) returns (SingleStringReply) {}
+  rpc QuestionAndAnswerReply(QARequest) returns (SingleStringReply) {}
+  rpc FillMaskReply(SingleStringRequest) returns (SingleStringReply) {}
+  rpc TokenClassificationReply(SingleStringRequest) returns (SingleStringReply) {}
+  rpc ConversationalReply(ConversationRequest) returns (ConversationReply) {}
+  rpc Txt2ImgReply(MultiStringRequest) returns (ImageReply) {}
```
mrwyattii, line 49: Why does the DeploymentManagement need to handle all of these?

mii/models/score/score_template.py
```diff
@@ -16,24 +16,41 @@
...
+        data = {
+            'deployment_name': deployment[mii.constants.DEPLOYMENT_NAME_KEY],
+            'task': deployment[mii.constants.TASK_NAME_KEY],
+            'model': deployment[mii.constants.MODEL_NAME_KEY],
+            'enable_deepspeed': deployment[mii.constants.ENABLE_DEEPSPEED_KEY],
+            'enable_zero': deployment[mii.constants.ENABLE_DEEPSPEED_ZERO_KEY],
+            'GPU_index_map': deployment[mii.constants.GPU_INDEX_KEY],
+            'mii_config': deployment[mii.constants.MII_CONFIGS_KEY],
+            'ds_config': deployment[mii.constants.DEEPSPEED_CONFIG_KEY],
+            'version': 1
```
mrwyattii, line 36: If we made the `mii.constants.*` values and the `DeploymentConfig` field names match (or add field aliases) we could just pass the entire `deployment` dict to `DeploymentConfig` rather than be extra verbose with the packing and unpacking of values.

mii/deployment.py
TosinSeg: I changed it and made a new function, let me know if you think it is better

mii/config.py
```diff
@@ -123,4 +125,23 @@ class LoadBalancerConfig(BaseModel):
...
+    deployment_name: str
+    task: str
+    model: str
+    enable_deepspeed: bool = True
+    enable_zero: bool = False
+    GPU_index_map: dict = None
+    mii_config: MIIConfig = MIIConfig.parse_obj({})
+    ds_config: dict = None
+    version: int = 1
+    deployed: bool = False
```
TosinSeg, line 143: Removed the deployed option or make it hidden _deployed(verify this is hidden)

mii/client.py
TosinSeg: Add new check to ensure that all the port numbers are the same

mii/config.py
```diff
@@ -123,4 +126,25 @@ class LoadBalancerConfig(BaseModel):
 
     class Config:
         validate_all = True
-        validate_assignment = True
+
+    validate_assignment = True
+
+
+class DeploymentConfig(BaseModel):
+    deployment_name: str = Field(alias="DEPLOYMENT_NAME_KEY")
```
mrwyattii, line 134: I'm a little bit confused about these aliases. I know we discussed a change here last week, but I think there may have been some miscommunication.

mii/client.py
```diff
@@ -55,51 +74,189 @@ class MIIClient():
...
+                   model=None,
+                   deployment_name=None,
+                   enable_deepspeed=True,
+                   enable_zero=False,
+                   ds_config=None,
+                   mii_config={},
+                   deployments=[],
+                   deployment_type=DeploymentType.LOCAL,
+                   model_path=None,
+                   version=1):
```
mrwyattii, line 184: Since this is a new interface, we do not need to support both the old and new way of adding models. Let's just allow passing of `deployments` and not `task, model, deployment_name`

mii/client.py
```diff
@@ -55,51 +74,189 @@ class MIIClient():
...
+            self.lb_stub = modelresponse_pb2_grpc.DeploymentManagementStub(channel)
+        self.lb_config = lb_config
+        self.model_path = model_path
+        self.port_map = port_map if port_map is not None else {}
+        self.deployment_tag = deployment_tag
+
+    async def add_models_async(self, proto_request):
+        await getattr(self.lb_stub, "AddDeployment")(proto_request)
+
+    def add_models(self,
```
mrwyattii, line 173: I'm a bit confused about the implementation of this method. It looks like we are re-generating the score file when we add models. And then we call `init` again? Does this not cause problems for models that are already running?

mii/client.py
```diff
@@ -55,51 +74,189 @@ class MIIClient():
...
+            self.lb_config.replica_configs.extend(lb_config.replica_configs)
+        else:
+            self.lb_config = lb_config
+        for deployment in deployments:
+            self.deployments[getattr(deployment,
+                                     mii.constants.DEPLOYMENT_NAME_KEY)] = deployment
+        if self.model_path is None and deployment_type == DeploymentType.LOCAL:
+            self.model_path = mii.constants.MII_MODEL_PATH_DEFAULT
+        elif self.model_path is None and deployment_type == DeploymentType.AML:
+            model_path = "model"
```
mrwyattii, line 219: We don't necessarily support multi-model / adding deployments with AML, so we should error out if deployment type is AML.

mii/client.py
```diff
@@ -55,51 +74,189 @@ class MIIClient():
...
+            self.channel = create_channel(host, port)
+            self.mr_stub = modelresponse_pb2_grpc.ModelResponseStub(self.channel)
+        self.deployments = deployments
+
+    def _get_deployment_task(self, deployment_name=None):
+        task = None
+        if deployment_name is None or deployment_name == mii.constants.MII_TERMINATE_DEP_KEY:  #mii.terminate() or single model
+            if deployment_name is None:
+                assert len(self.deployments) == 1, "Must pass deployment_name to query when using multiple deployments"
+            deployment = next(iter(self.deployments.values()))
```
mrwyattii, line 92:
```suggestion
            deployment = self.deployments.values()[0]
```

PR #213, opened 2023-07-25
Add Pydantic v2 support

mrwyattii:


ringohoffman:
From the [migration guide](https://docs.pydantic.dev/latest/migration/):

> Due to performance overhead and implementation complexity, we have now deprecated support for specifying json_encoders in the model config. This functionality was originally added for the purpose of achieving custom serialization logic, and we think the new serialization decorators are a better choice in most common scenarios.

loadams:
Closing this instance as it is repetitive with the actual pydantic PR that will be merged

PR #211, opened 2023-07-21
Update AML Deployment

mrwyattii:
Recent feature additions like load balanced model replicas were not properly tested on the AML deployments. Updating the AML deployments and making minor improvements/fixes to work with multiple model replicas.

Co-Author: @DannyBruno
Co-Author: @srsaggam

PR #214, opened 2023-07-30
hack: add llama2 models

stillmatic:
these models have the correct tags in HF but do not show up for some reason here.

Note that these require having the HF Hub token and an authorized HF account / signing Llama TOS. I also have not included the 70b model - not sure if DeepSpeed has implemented GQA yet.

I have signed the Microsoft CLA

### DIFF
--- a/mii/config.py
+++ b/mii/config.py
@@ -3,7 +3,7 @@
 
 # DeepSpeed Team
 import torch
-from typing import Union, List
+from typing import Optional, Union, List
 from enum import Enum
 from pydantic import BaseModel, validator, root_validator
 
@@ -55,7 +55,7 @@ class MIIConfig(BaseModel):
     enable_restful_api: bool = False
     restful_api_port: int = 51080
     replica_num: int = 1
-    hostfile: str = DLTS_HOSTFILE
+    hostfile: Optional[str] = DLTS_HOSTFILE
     trust_remote_code: bool = False
 
     @validator("deploy_rank")
--- a/mii/deployment.py
+++ b/mii/deployment.py
@@ -107,9 +107,9 @@ def deploy(task,
     if mii_config.hostfile is None:
         hostfile = tempfile.NamedTemporaryFile(delete=False)
         num_gpu = torch.cuda.device_count()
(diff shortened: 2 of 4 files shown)


mrwyattii:
@stillmatic what error is this resolving? I'm able to launch llama2 models with the current `main` branch of DeepSpeed-MII:
```python
import mii
mii_config = {"dtype":"fp16", "hf_auth_token":<YOUR_HF_TOKEN>}
mii.deploy("text-generation", model="meta-llama/Llama-2-7b-hf", deployment_name="llama-deployment", mii_config=mii_config)
generator = mii.mii_query_handle("llama-deployment")
result = generator.query({"query": "test one two"}, max_new_tokens=1024)
```

It looks like you are modifying AML deployment code. Are you trying to get this working on AML resources? If so, we are aware of bugs with the AML deployment type and are resolving them in this PR: #211

stillmatic:
cool -- I think issue was with the pip release, confirm master works (for moving past the 'unsupported model' errors) - closing the PR. 

re the deployment code: I'm running locally in single-gpu setting and hitting this error

```
> mii.deploy("text-generation", model="meta-llama/Llama-2-7b-hf", deployment_name="llama-deployment", mii_config=mii_config)

    176 def _allocate_processes(hostfile_path, tensor_parallel, num_replicas):
    177     resource_pool = fetch_hostfile(hostfile_path)
--> 178     assert resource_pool is not None and len(
    179         resource_pool) > 0, f'No hosts found in {hostfile_path}'
    181     replica_pool = []
    182     allocated_num = 0

AssertionError: No hosts found in /job/hostfile
```

so was trying to work around that.

mrwyattii:
@stillmatic this looks like a bug on our side. We should be auto-generating a hostfile if one is not found, but it looks like that is not happening. You can avoid this by creating a file at `/job/hostfile` with the following content:
`localhost slots=<NUMBER OF GPUS>`

For example, on my local machine I have 2 A6000s and use:
`localhost slots=2`

I've created an issue and will follow up soon with a fix: #216

PR #218, opened 2023-08-04
Refactor Configs

mrwyattii:
In order to better support multi-model deployments and associated features from #208 we have identified some key ways in which the configs can be refactored. This will be the first of 2-3 PRs that does some refactor work and implements smaller segments of the work done in #208 to make review and merging easier.

@TosinSeg 

Work done in this PR:
- Refactor our configs to have `MIIConfig` and `DeploymentConfig`. This is in anticipation of having multi-model deployments where we will be passing in multiple `DeploymentConfig` objects
  - This could probably be extended in this PR to also include the DeepSpeed `InferenceConfig` rather than recreating that config in `load_models.py` (see comments I left on the code changes)
- With the refactoring of our configs, we can simplify a lot of our code:
  - Avoid packing/unpacking config data at various stages of the deployment. Instead, just pass around the `mii_config` object as it contains everything we want. This massively simplifies the score template and server class
  - Move lots of parameter and config verification that was scattered around the codebase into the config classes themselves. This consolidates code to a common location and simplifies the process of debugging config settings
- Light formatting changes. We use `yapf` for formatting, but this allows mixed `'` and `"` - I don't think we need to use one or the other, but being consistent is good. Applied `black` and then `yapf` to get some nicer formatting (especially on the long assert lines)
  - Is it too late to switch the project to use `black`? @jeffra 

I left lots of comments about changes I felt unsure about or where I could use some feedback, however the PR is mostly ready to merge. We could always handle minor cleanup and changes in future PRs that implement the rest of #208


mii/client.py
```diff
@@ -39,27 +35,33 @@ def mii_query_handle(deployment_name):
         inference_pipeline, task = mii.non_persistent_models[deployment_name]
         return MIINonPersistentClient(task, deployment_name)
 
-    task_name, mii_configs = _get_deployment_info(deployment_name)
-    return MIIClient(task_name, "localhost", mii_configs.port_number)
+    mii_config = _get_mii_config(deployment_name)
+    return MIIClient(mii_config.deployment_config.task,
```
mrwyattii, line 39: Should probably make this just accept mii_config. Is `"localhost"` ever a different value?

mii/client.py
```diff
@@ -155,30 +161,32 @@ def destroy_session(self, session_id):
...
+        assert (
+            self.deployment_name in mii.non_persistent_models
+        ), f"deployment: {self.deployment_name} not found"
         task_methods = GRPC_METHOD_TABLE[self.task]
         inference_pipeline = mii.non_persistent_models[self.deployment_name][0]
 
-        if self.task == Tasks.QUESTION_ANSWERING:
-            if 'question' not in request_dict or 'context' not in request_dict:
+        if self.task == TaskType.QUESTION_ANSWERING:
+            if "question" not in request_dict or "context" not in request_dict:
```
mrwyattii, line 177: We should use this dict to check if the required keys are present:
https://github.com/microsoft/DeepSpeed-MII/blob/b0f3f650290e3d9d34e0d4b10141ba5f5d6bb67c/mii/constants.py#L44

mii/config.py
```diff
@@ -3,93 +3,147 @@
...
+    task: TaskType
+    dtype: DtypeEnum = DtypeEnum.fp32
+    model_path: str = ""
+    checkpoint_dict: Optional[Dict[str, Any]] = None
+    max_tokens: int = 1024
 
-        # ensure deploy rank type is always list for easier consumption later
-        if not isinstance(field_value, list):
-            field_value = [field_value]
+    # Performance configs
```
mrwyattii, line 47: Do we want to use DeepSpeed's `InferenceConfig` here? This would expand on the options that we expose to users and we could pass the config directly to DeepSpeed in `load_models.py`!

mii/config.py
```diff
@@ -3,93 +3,147 @@
...
-    meta_tensor: bool = False
+
+class ReplicaConfig(DeepSpeedConfigModel):
+    hostname: str = ""
+    tensor_parallel_ports: List[int] = []
+    torch_dist_port: int = None
+    gpu_indices: List[int] = []
+
+
+class DeploymentConfig(DeepSpeedConfigModel):
```
mrwyattii, line 25: Is this the organization of config params we want? Is it worth having model/inferences configs be sub-configs of the deployment config?

mii/config.py
```diff
@@ -3,93 +3,147 @@
...
+    tensor_parallel: int = 1
+    enable_cuda_graph: bool = False
+    replace_with_kernel_inject: bool = True
 
-        # number of ranks provided must be equal to TP size, DP is handled outside MII currently
-        assert values["tensor_parallel"] == len(field_value), \
-            f"{len(field_value)} rank(s) provided in 'deploy_rank' does not align with tensor_parallel size of {values['tensor_parallel']}"
-        return field_value
+    class Config:
+        json_encoders = {torch.dtype: lambda x: str(x)}
```
mrwyattii, line 56: We may want to bring this change into the `DeepSpeedConfigModel`

mii/config.py
```diff
@@ -99,29 +153,124 @@ def meta_tensor_or_sys_mem(cls, values):
...
-class ReplicaConfig(BaseModel):
-    hostname: str = ""
-    tensor_parallel_ports: List[int] = []
-    torch_dist_port: int = None
-    gpu_indices: List[int] = []
 
-    class Config:
-        validate_all = True
-        validate_assignment = True
+class MIIConfig(DeepSpeedConfigModel):
```
mrwyattii, line 177: TODO: Bring the comments on what each option does (from `deployment.py`) to here. Make compatible with the auto-generated pydantic docs (and create the docs for MII)

mii/config.py
```diff
@@ -99,29 +153,124 @@ def meta_tensor_or_sys_mem(cls, values):
...
+            assert (
+                set(values.get("deployment_config").deployment_name) <= allowed_chars
+            ), "AML deployment names can only contain a-z, A-Z, 0-9, and '-'."
+        return values
 
-    class Config:
-        validate_all = True
-        validate_assignment = True
+    @root_validator(skip_on_failure=True)
+    def generate_replica_configs(cls, values):
```
mrwyattii, line 207: This validator is not great. Need to find a cleaner way to setup the replica configs. Also, we may not want to do replica config construction when we call to `mii.deploy()`. In the case that a user is not spinning up MII on the machine they run `deploy()` (e.g., with AML deployment), then this will cause errors because it may not find the correct number of GPUs :(

mii/deployment.py
```diff
@@ -2,94 +2,70 @@
...
+            deployment_config[key] = val
+
+    mii_config = {k: v for k, v in mii_config.items() if hasattr(MIIConfig, k)}
+    mii_config["version"] = version
+    mii_config["deployment_type"] = deployment_type
+
+    return deployment_config, mii_config
+
+
+def deploy(deployment_name: str,
```
mrwyattii, line 49: Would it be better to have a different API? Trying to think of the multi-model deployment case when designing the inputs. Imagine a multimodel setup:
```python
mii_config = {"port_number": 50000}
deployment_configs = [{"task":"text-generation", "model":"gpt2"}, {"task":"text-generation", "model":"bigscience/bloom-560m"}]
deployment_name = "my-multi-model-dep"
mii.deploy(deployment_name, deploymetn_configs, mii_config)
```

Can we find something better?

mii/deployment.py
```diff
@@ -2,94 +2,70 @@
...
-    Returns:
-        If deployment_type is `LOCAL`, returns just the name of the deployment that can be used to create a query handle using `mii.mii_query_handle(deployment_name)`
-
-    """
-
-    # parse and validate mii config
+from .config import MIIConfig, DeploymentType
+
+
+def support_legacy_api(
```
mrwyattii, line 14: This should support all cases of the legacy API - however we may want to add a unit test. It depends if we plan to deprecate it soon or keep the old API around.

mii/deployment.py
```diff
@@ -98,112 +74,39 @@ def deploy(task,
...
-                                                      task)
-    else:
-        raise Exception(f"Unknown deployment type: {deployment_type}")
+    if mii_config.deployment_type != DeploymentType.NON_PERSISTENT:
+        create_score_file(mii_config)
+
+    if mii_config.deployment_type == DeploymentType.AML:
+        _deploy_aml(mii_config)
+    elif mii_config.deployment_type == DeploymentType.LOCAL:
+        return _deploy_local(mii_config)
```
mrwyattii, line 83: I think this is a pointless `return` - should probably just call `_deploy_local`

mii/deployment.py
```diff
@@ -98,112 +74,39 @@ def deploy(task,
...
+        assert (
+            int(os.getenv("WORLD_SIZE", "1"))
+            == mii_config.deployment_config.tensor_parallel
+        ), "World Size does not equal number of tensors. When using non-persistent deployment type, please launch with `deepspeed --num_gpus <tensor_parallel>`"
+        deployment_name = mii_config.deployment_config.deployment_name
+        task = mii_config.deployment_config.task
+        mii.non_persistent_models[deployment_name] = (
+            load_models(deployment_config),
+            task,
+        )
```
mrwyattii, line 94: Move this to a `_deploy_nonpersistent()` to match the other deployment modes

mii/models/load_models.py
```diff
@@ -64,44 +58,42 @@ def load_models(task_name,
...
-                                                mii_config)
-        inf_config["replace_with_kernel_inject"] = False  #not supported yet
+
+        inference_pipeline = diffusers_provider(deployment_config)
         inf_config["enable_cuda_graph"] = True
     else:
         raise ValueError(f"Unknown model provider {provider}")
-
+    """
     print(
```
mrwyattii, line 70: Need to re-enable this print. I think it's important to give these snippets of output for users to understand the progress of MII deployment (and see what optimizations are being applies). However, I'm wondering if this should be moved elsewhere, maybe a utils function that takes in `mii_config` or `deployment_config` and prints useful things?

mii/models/load_models.py
```diff
@@ -64,44 +58,42 @@ def load_models(task_name,
...
-                                                task_name,
-                                                mii_config)
-        inf_config["replace_with_kernel_inject"] = False  #not supported yet
+
+        inference_pipeline = diffusers_provider(deployment_config)
         inf_config["enable_cuda_graph"] = True
     else:
         raise ValueError(f"Unknown model provider {provider}")
-
+    """
```
mrwyattii, line 69: Need to re-enable this print. I think it's important to give these snippets of output for users to understand the progress of MII deployment (and see what optimizations are being applies). However, I'm wondering if this should be moved elsewhere, maybe a utils function that takes in mii_config or deployment_config and prints useful things?

mii/server.py
```diff
@@ -102,242 +80,81 @@ def _is_server_process_alive(self, process):
...
+        ]
 
         host_gpus = defaultdict(list)
-        for repl_config in lb_config.replica_configs:
+        for repl_config in mii_config.deployment_config.replica_configs:
             host_gpus[repl_config.hostname].extend(repl_config.gpu_indices)
 
         # Start replica instances
-        for i, repl_config in enumerate(lb_config.replica_configs):
+        for repl_config in mii_config.deployment_config.replica_configs:
```
mrwyattii, line 127: I refactored this to remove the `_launch_{X|Y|Z}()` methods - but this this function got a bit long in the process. It may be cleaner to move back to the individual functions.

mii/config.py
```diff
@@ -3,93 +3,147 @@
...
+
+class DeploymentConfig(DeepSpeedConfigModel):
+    # Deployment configs
+    deployment_name: str
     load_with_sys_mem: bool = False
-    enable_cuda_graph: bool = False
-    checkpoint_dict: Union[dict, None] = None
-    deploy_rank: Union[int, List[int]] = -1
+    meta_tensor: bool = False
+    hf_auth_token: str = ""
```
mrwyattii, line 30: Make this more general like a `pipeline_kwargs` that all get passed to `transformers.pipeline()`

jeffra (approved):

mii/deployment.py
tohtana: In my opinion, we should stop supporting old APIs as soon as possible. We don't have many APIs and they are not closely related. It should be easy for users to migrate to new ones.

tohtana (approved):
Thank you for this great effort! I've noticed that the areas I had concerns about have been improved.  This will make future development much more pleasant.

PR #267, opened 2023-11-07
set device on inference pipeline only if setter available

gauravrajguru:
Diffuser pipeline don't have device setter method. Added error handling for it.


mii/legacy/models/load_models.py
```diff
@@ -89,7 +89,9 @@ def load_models(model_config):
         ds_engine.module.eval()  # inference
         inference_pipeline.model = ds_engine.module
 
-    if model_config.load_with_sys_mem:
+    # Set device property only if setter method available.
+    if (model_config.load_with_sys_mem
```
mrwyattii, line 93: What was the deployment setup that caused this to error out? I'm trying to understand if this is the proper solution. If there is no device property for the inference pipeline, then perhaps we need to set device a different way for the model?

gauravrajguru: I'm getting this error even with diffusers==0.15 library. 

For diffuser pipeline, device property is there, but setter is not defined for this only getter is present.

In transformer library's pipelines have device as class member. However, diffuser pipeline (from diffuser library) have device property only getter defined and the setter. Hence, getting following exception stack. 

I believe, the device is assigned to diffuser pipeline by using device_map parameter in from_pretrained function, because device property already has correct value.

Stack:
Traceback (most recent call last):
  File "/opt/conda/envs/ptca/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/conda/envs/ptca/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/mii/launch/multi_gpu_server.py", line 97, in <module>
    main()
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/mii/launch/multi_gpu_server.py", line 89, in main
    inference_pipeline = load_models(args.model_config)
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/mii/models/load_models.py", line 93, in load_models
    inference_pipeline.device = torch.device(f"cuda:{local_rank}")
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/diffusers/pipelines/pipeline_utils.py", line 522, in __setattr__
    super().__setattr__(name, value)
AttributeError: can't set attribute 'device'

I see your point, should we use to() method to set device like this pipe = pipe.to("cuda")
https://github.com/huggingface/diffusers/blob/e7534542a2e736ab54328a7fb3a0a15fe4f31da2/src/diffusers/pipelines/pipeline_utils.py#L599

mrwyattii:
Closing in favor of #279

PR #268, opened 2023-11-07
Expose top-p, top-k, and temperature to generate APIs

mrwyattii:
I reworked how we do batch processing, such that we can utilize multiple logit post processing techniques for a single query. In doing so, I've also added a cache of post processing objects that is stored in the Pipeline objects.

TODO:

- [x] Verify no impact on performance
- [x] Add unit tests
- [x] Update docs


mii/batching/ragged_batching.py
```diff
@@ -536,51 +535,62 @@ def make_request(self,
...
-            SAMPLERS)
-        stop_criterion = _create_postprocessor(
-            postprocess_config.get("stop_criterion",
-                                   DEFAULT_STOP_CRITERION),
-            STOP_CRITERIA,
-            {"tokenizer": self.tokenizer},
-        )
+        post_processing = []
+
+        top_p = kwargs.pop("top_p", 0.9)
```
tohtana, line 550: The constants (keys and value) can be defined somewhere else, like constans.py.

mii/batching/ragged_batching.py
```diff
@@ -536,51 +535,62 @@ def make_request(self,
...
+        stop_name = "Stop"
+        if stop_name not in self._post_processors:
+            self._post_processors[stop_name] = EosGenerationStopCriterion(
+                tokenizer=self.tokenizer)
+        post_processing.append(stop_name)
 
         assert kwargs == {}, f"Unknown keyword arguments {kwargs}"
 
         return [
             RaggedRequest(
```
tohtana, line 585: Just for sharing information, we now decompose prompts in the scheduler with the dynamic decomposition (SplitFuse). So it doesn't make sense to return this `RaggedRequest` as a list. We can simplify this in another refactoring PR.

tohtana (approved):
This PR overall looks good to me. I left a few comments in terms of refactoring but we could do it later according to our priority.

PR #269, opened 2023-11-08
Reorganize code structure, fix client import bug

mrwyattii:
This PR is a first step to better organize the MII code. Major changes include:
- Creating an `api.py` file for user-facing functions
- Moving `client.py` and `server.py` into `backend/` (this also resolves a bug where we previously did `from .client import client` in our `__init__.py` file, which caused the `mii.client` submodule to be replaced with the `mii.client` function)
- Moving `models.py` and `tokenizers.py` into `modeling/`
- Moving `task_methods.py` under `grpc_related/` (Because it is only used with the persistent GRPC deployment)

@jeffra


jeffra (approved):
This is great, thank you @mrwyattii. @tohtana please take a look as well but otherwise looks good to me.

mii/api.py
```diff
@@ -0,0 +1,149 @@
...
+          mii_config: Optional[Dict[str,
+                                    Any]] = None,
+          **kwargs) -> Union[None,
+                             MIIClient]:
+    mii_config = _parse_kwargs_to_mii_config(model_name_or_path=model_name_or_path,
+                                             model_config=model_config,
+                                             mii_config=mii_config,
+                                             **kwargs)
+
+    #MIIServer(mii_config)
```
tohtana, line 103: Can we remove this?

mii/api.py
```diff
@@ -0,0 +1,149 @@
...
+            model_name=mii_config.model_config.model,
+            task_name=mii_config.model_config.task,
+            replica_num=mii_config.model_config.replica_num,
+            instance_type=mii_config.instance_type,
+            version=mii_config.version,
+        )
+        print(
+            f"AML deployment assets at {mii.aml_related.utils.aml_output_path(mii_config.deployment_name)}"
+        )
+        print("Please run 'deploy.sh' to bring your deployment online")
```
tohtana, line 123: It would be good to have a recommended method to output the log, logger or print.
We may not need to work on it in this PR, let me just share the idea before I forget.

tohtana (approved):
It looks good to me. I left small comments to share my thought, but I don't think we need to address them in this PR

mii/api.py
mrwyattii: Let me update this explain why I'm keeping it here!

PR #274, opened 2023-11-08
Mitigate the risk of deadlock

tohtana:
This PR includes:
- Display message when we fall into the deadlock caused by inefficient KV cache
- Keep some margin when scheduling requests


mrwyattii (approved):

cmikeh2 (approved):
I don't know if I'm sufficient to approve a review here for merge, but I like this 😄

PR #278, opened 2023-11-09
Recompute when the deadlock is detected

tohtana:
Wit this PR, the scheduler clears the KV cache and recomputes it when the deadlock is detected.


mii/batching/ragged_batching.py
```diff
@@ -147,6 +147,10 @@ class RaggedRequest:
     _generated_tokens: List[torch.Tensor] = field(default_factory=list)
     _finish_reason: GenerationFinishReason = GenerationFinishReason.NONE
 
+    @property
+    def prompt_length(self) -> int:
```
mrwyattii, line 151: In the case we have deadlock, the `prompt_length` will be incorrect on the output. Perhaps we need to store `original_prompt_length` or something like that. Don't worry about this for now. I have plans to expand the output details and can handle this in a future PR.

mrwyattii (approved):

mii/batching/ragged_batching.py
tohtana: I see, you are right. I would appreciate it if you could fix it on your PR

mrwyattii (approved):

Issue #290, opened 2023-11-13
Why did an unknown error occur in gprc calls when enable_restful_api is true 


wtontheway:
I don't know the reason why this error occurred。After started the deepspeed service using the following code， using the curl command to call the restful service, then the error occurred。

```
import mii
from mii import DeploymentType

mii_configs = {
    "deployment_type": DeploymentType.LOCAL,
    "enable_restful_api": True,
    "deployment_name": "zzz-mii",

}

model_configs = {
    "replica_num": 1,
    "tensor_parallel": 1,
}

mii.serve("/mnt/home/LlaMa2/file",
          model_config=model_configs,
          mii_config=mii_configs, )
```

```
root@ctm:/mnt/home/zzz/mii-test# curl  -X POST  -H "Content-Type: application/json" -d '{"request": {"query": "DeepSpeed is the greatest"}}' http://localhost:51080/mii/zzz-mii
[2023-11-13 15:46:45,323] ERROR in app: Exception on /mii/zzz-mii [POST]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 867, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 852, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/local/lib/python3.8/dist-packages/flask_restful/__init__.py", line 489, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask/views.py", line 109, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask_restful/__init__.py", line 604, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mii/grpc_related/restful_gateway.py", line 31, in post
    result = client.query(data["request"], **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mii/legacy/client.py", line 77, in query
    return self.asyncio_loop.run_until_complete(
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/usr/local/lib/python3.8/dist-packages/mii/legacy/client.py", line 73, in _request_async_response
[...]

PR #294, opened 2023-11-13
Update RESTful API

### LINKED ISSUE #270: Support AsyncPipeline for RESTful API
Are you planning to support this feature?
I'm wanna use FastGen in my app but it's not currently support RESTful API asynchronously
vLLM support it's very well: https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/api_server.py

I've also use Ray to deploy a server use MIIPipeline with dynamic batching but the performance is far behind vLLM default settings.

mrwyattii:
Bringing this feature back from MII-Legacy. Resolves #270 and #290


lekurile (approved):
LGTM!

PR #295, opened 2023-11-13
Unify input/output types

mrwyattii:
Several changes in this PR, with the main goal to make the inputs/outputs of `pipeline` and `client.generate` match. That is, each can take `str|List[str]` as input and will return `List[Response]` where `Response` is `mii.batching.data_classes.Response`.

In addition, I've modified the entire chain of communication between the MII user-facing APIs to the inference server to better represent the inputs and outputs. This includes removing many protobuf message types that are using in MII-Legacy, but not the latest MII.

Full summary of changes:
- Move `Response`, `Request`, etc. classes to new `data_classes.py` file
- Made the output of `client.generate` match that of `pipeline`
- Removed old GRPC/proto code that is unused
- Changed how we represent response messages to better capture generation details

_Note: This looks like a lot of change, but it's mostly moving things around the recompiling the proto stuff_

TODO:

- [x] Test pipeline
- [x] Performance testing
- [x] Docs
- [x] Update DS and DSExamples (No changes for DS, https://github.com/microsoft/DeepSpeedExamples/pull/812)


mii/batching/data_classes.py
```diff
@@ -0,0 +1,236 @@
...
+    finish_reason: GenerationFinishReason
+
+    @staticmethod
+    def from_msg_dict(msg: Dict[str, Union[str, int]]) -> Self:
+        return Response(**msg)
+
+    def to_msg_dict(self) -> Dict[str, Union[str, int]]:
+        return asdict(self)
+
+    def __repr__(self) -> str:
```
tohtana, line 28: Can we show other attributes including finish_reason?

tohtana (approved):
This achievement is great, thank you!
Since we don't have `response.response` any more in the client code, we need to make sure we update the benchmark scripts in DeepSpeedExamples after this PR is merged or the package on PyPI is updated

jeffra (approved):

PR #311, opened 2023-11-16
readable token streaming support

jeffra:
## Example code
```python
import mii

out_tokens = []
def callback(response):
    print(f"recv: {response.response[0]}")
    out_tokens.append(response.response[0])
prompt = "Washington state is in the PNW"

print(f"{prompt=}")
client = mii.serve("meta-llama/Llama-2-7b-hf")
client.generate(prompt, max_new_tokens=20, streaming_fn=callback)
final_str = prompt + "".join(out_tokens)
print(f"{final_str=}")
```

## Results
```bash
prompt='Washington state is in the PNW'
recv: .
recv:  We
recv: ather
recv:  here
recv:  var
recv: ies
recv:  from
recv:  north
recv:  to
recv:  south
recv: ,
recv:  from
recv:  west
recv:  to
recv:  east
recv: .
recv:  The
recv:  closer
recv:  to
recv:  the
recv: 
final_str='Washington state is in the PNW. Weather here varies from north to south, from west to east. The closer to the'
````

- [ ] add unit test for `ReadableStream` class
- [ ] triage fixes for this unrelated issue: https://github.com/microsoft/DeepSpeed/pull/4668/files#r1395042852

Should address: https://github.com/microsoft/DeepSpeed-MII/issues/306, https://github.com/microsoft/DeepSpeed-MII/issues/281

### DIFF
--- a/mii/batching/ragged_batching.py
+++ b/mii/batching/ragged_batching.py
@@ -477,13 +477,12 @@ class RaggedBatchBase:
     def _generate_output(self, r: RaggedRequest) -> bool:
         outputs = []
         if r.stream:
-            outputs.append((
-                r.uid,
-                [r.next_token],
-                r.prompt_length,
-                r.num_generated_tokens,
-                GenerationFinishReason.NONE,
-            ))
+            outputs.append((r.uid,
+                            [r.next_token],
+                            r.prompt_length,
+                            r.num_generated_tokens,
+                            GenerationFinishReason.NONE,
+                            r.stream))
         if r.finish_reason != GenerationFinishReason.NONE:
             if r.stream or not r.generated_tokens:
                 output_tokens = []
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: mii/batching/ragged_batching.py [hunk 1]
```diff
@@ -721,6 +719,44 @@ def flush(self, uids: List[int]) -> None:
             self.inference_engine.flush(uid)
 
 
+@dataclass
+class StreamState:
+    prev_tok: str
+    tids: List[int]
```
mrwyattii, line 725: This might be easily confused with the thread ID variable `tid`.  Could we use `token_ids` or replace `tid` with `thread_id` in the `get_response` method?

Occupying-Mars:
Has this been implemented in the main branch and if not is there any way to stream on the current mii package only

spring1915:
How to relay the stream to the server API? For instance, I use [djl-python server](https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/deepspeed.py#L512) with `Input` and `Output`:
```
def handle(inputs: Input) -> Optional[Output]:
    if not _service.initialized:
        _service.initialize(inputs.get_properties())

    if inputs.is_empty():
        return None

    return _service.inference(inputs)
```
How to send the streaming generated tokens to the client?

mrwyattii:
Closing in favor of #397

PR #359, opened 2023-12-20
Add pipeline unit tests

mrwyattii:
- Add `destroy()` method to pipeline objects
- Add unit tests for pipeline
- Fix kwarg parsing bug, where passing a `model_config` inside a `mii_config` resulted in an empty `model_config`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,7 +5,6 @@
 
 import pytest
 import time
-import torch
 import os
 import mii
 from types import SimpleNamespace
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -156,6 +152,7 @@ def deployment(mii_config, expected_failure):
             mii.serve(mii_config=mii_config)
         yield excinfo
     else:
+        print("CONFIG", mii_config)
```
tohtana, line 155: Do we need to keep this print?

tohtana (approved):

tests/conftest.py
mrwyattii: We do not, removing this

mrwyattii, line 155:
```suggestion
```

PR #373, opened 2024-01-11
Add model support unit test

mrwyattii:
Testing a model from each of the support model families. These tests create a random weights checkpoint engine and inject into deepspeed. This saves time running the tests and we only test if they generate output. These are not correctness tests.


arashb (approved):
LGTM 👍

PR #375, opened 2024-01-12
Remove inefficient loop in TopP logits processor

tohtana:
The loop in Top P processor is inefficient when it runs on multiple requests.
This PR replaces the loop with a batch API with indexing.


tohtana:
Closed because an inappropriate commit is included

PR #376, opened 2024-01-12
Improve efficiency of ragged batching scheduler

tohtana:
This PR includes the following optimization in the ragged batching scheduler:
- Remove inefficient loop in TopP logits processor (390302476251c9d58ea443900a0388e8d006f19f): The loop in Top P processor is inefficient when it runs on multiple requests. This PR replaces the loop with a batch API with indexing. 
- Copying `is_done` flag of requests to host memory as a batch (678673734503b14ed5a9b9dd7ec7d7e1efa6d0eb): is_done is computed on cuda but implicitly copied to host every time it is evaluated as bool. This PR copies it to host at once and stores them as Python bool values.

### DIFF
--- a/mii/batching/data_classes.py
+++ b/mii/batching/data_classes.py
@@ -225,20 +225,20 @@ class RequestBatch:
         return [r.next_token for r in self.requests]
 
     @property
-    def done_tokens(self) -> List[torch.Tensor]:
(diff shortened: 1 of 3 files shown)


tohtana:
Closing because #377 is an improved version of this PR.

Issue #347, opened 2023-12-05
How to stream tokens?


mevince:

mrwyattii:
Hi @mevince we are currently working on bring proper streaming support in #311. You can try out that branch with `pip install git+https://github.com/Microsoft/DeepSpeed-MII@jeffra/readable-streaming`. There is example code in the description for how to do token streaming.

Issue #306, opened 2023-11-15
Support for token streaming


Archmilio:
Thank you for your hard work.
I am really excited about MII performance.

I have some questions

Does token streaming function supported now?

If token streaming is supported, I would like to test the first token latency and completion time. Do you happen to know when it will be supported?

Issue #281, opened 2023-11-10
Streaming api seems broken


ttim:
Currently while using streaming API output seems off - there is no spaces in response and it's unclear where to add them.
I've tried printing in `streaming_fn` following strings:
- `"".join(item.response)` in this case there is no spaces in output
- `" ".join(item.response)` in this case still no spaces
- `"".join(item.response) + " "` - too many spaces, seems like after each token

I'm not sure if it's something wrong I do or something is broken, any help would be great!

PawanOsman:
I have the same issue

mrwyattii:
Hi @ttim and @PawanOsman we have not officially released the streaming API for this reason. @jeffra is actively working to fix these problems and you can expect an update and official release of this feature very soon!

PR #397, opened 2024-01-30
Readable token streaming support

greshilov:
Hey folks,

Thank you for this amazing library! :)  

I waited a bit for #311 to resolve, but eventually decided to help complete it.  

The original @jeffra's commit is cherry-picked, so the commit history stays valid!

# Description

While testing the original PR, I stumbled upon several issues:

* It trimmed redundant spaces (`the   example` becomes `the example`), because it was based on the space splitting

* It sometimes replaced more characters than needed (if the previous token was found multiple times in the current token, like `an anchor` becomes `an chor`)

* It didn't work properly when more than two tokens are required to decode the symbol

For example, with `tiiuae/falcon-7b` tokenizer, `例如` becomes `�如` (the first symbol is a Unicode `REPLACEMENT CHARACTER`)  

My approach is slightly different then used before.

Two consequent tokens are always stored in the form of a token_ids sequence. The idea is to find the difference between previous and (previous + current) token decode results:

```python
def example(...)
    ...
    # n - is the size of the previous token
    prev = tokenizer.decode(token_ids[:n])
    both = tokenizer.decode(token_ids)
    result = both.replace(prev, "", 1)
    token_ids = token_ids[n:]
    return result
```

I added unit tests for the `ReadableStream` for all supported models + one sanity test checking consistency between regular inference and streaming inference.

Should address: #306, #281, #347

### DIFF
--- a/mii/batching/ragged_batching.py
+++ b/mii/batching/ragged_batching.py
@@ -9,6 +9,7 @@ import queue
 import random
 import threading
 import time
+from dataclasses import dataclass
 from collections import deque, defaultdict
 from functools import cached_property
 from typing import Dict, Tuple, List, Any, Union, DefaultDict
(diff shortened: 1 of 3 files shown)


greshilov:
@microsoft-github-policy-service agree company="JetBrains"

mrwyattii (approved):

mrwyattii:
This is amazing! Thank you for the contribution @greshilov

PR #403, opened 2024-02-02
Improve recovery from KV cache starvation

tohtana:
This PRs improves efficiency when we are running out of KV cache.

Currently, we free the KV cache for all ongoing requests when we fail to schedule any request. This PR changes this strategy by only freeing the cache allocated to the last request. This approach reduces the recomputation cost and prevents the generation from getting stuck.

### DIFF
--- a/mii/batching/ragged_batching.py
+++ b/mii/batching/ragged_batching.py
@@ -269,9 +269,10 @@ class RaggedBatchBase:
 
             max_blocks = free_blocks - self.scheduled_req_blocks
 
-            # Check capacity to mitigate the deadlock risk
-            # We don't schedule requests when we find that a prompt is too long to fit to the KV cache
             if len(r.input_tokens) > 1:
+                # When the KV cache is out of capacity, we release KV cache blocks for a request.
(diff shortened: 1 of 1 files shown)


mii/batching/ragged_batching.py
```diff
@@ -349,20 +347,45 @@ def _queue_flush_request(self, uid: int) -> None:
             ))
 
     def reset_request_status(self):
+        ## Get the last request that consumes KV cache
+        last_r = None
         for r in self.buffer:
             if r.seq_length > 0:
-                self._queue_flush_request(r.uid)
+                last_r = r
+        assert last_r is not None
```
mrwyattii, line 355: Can we add an error message here to indicate something has gone wrong?

mrwyattii (approved):

mii/batching/ragged_batching.py
tohtana: Thank you for the comment! I added a message

mrwyattii:
CI: https://github.com/microsoft/DeepSpeed/actions/runs/7761886259

PR #421, opened 2024-02-23
Remove references to --extra-index-url in MII repo

loadams:


lekurile (approved):

mii/aml_related/templates.py
```diff
@@ -162,7 +162,7 @@
     CONDA_DEFAULT_ENV=amlenv \
     PATH=$PATH:/usr/local/cuda/bin
 
-RUN /opt/miniconda/envs/amlenv/bin/pip install -r "$BUILD_DIR/requirements.txt" --extra-index-url https://download.pytorch.org/whl/cu113 && \
+RUN /opt/miniconda/envs/amlenv/bin/pip install -r "$BUILD_DIR/requirements.txt" --index-url https://download.pytorch.org/whl/cu113 && \
```
loadams, line 165: Check with @mrwyattii on if this file is still used, since otherwise we are unlikely to be able to download the other packages in requirements.txt other than torch from here.  We may need to split this into 2 calls, install torch first and then the requirements.txt?

PR #470, opened 2024-05-06
fix max_ragged_sequence_count check in _schedule_prompts

dc3671:
- scheduling prompt should break when length of request **equals** `max_ragged_sequence_count`
- considering there are flush requests, should use `scheduled_requests.requests_to_run` rather than `scheduled_requests`

@loadams @mrwyattii can you help review this fix? thanks~


dc3671:
@loadams Can you trigger the CI again? Thanks

tohtana (approved):
Thank you @dc3671 for your contributions!

Issue #489, opened 2024-06-05
[BUG] MII Backend Hangs After 9999 Exceptions in `MIIAsyncPipeline.put_request`


weiqisun:
MII invokes [`MIIAsyncPipeline`](https://github.com/microsoft/DeepSpeed-MII/blob/c171c4ee290e96c0d3e618b654be8add5eca973b/mii/batching/ragged_batching.py#L635) for persistent deployments. During runtime, requests are passed from `GeneratorReply` to the backend model [through](https://github.com/microsoft/DeepSpeed-MII/blob/c171c4ee290e96c0d3e618b654be8add5eca973b/mii/grpc_related/modelresponse_server.py#L73) the `MIIAsyncPipeline.put_request` function. In this function, it [requests](https://github.com/microsoft/DeepSpeed-MII/blob/c171c4ee290e96c0d3e618b654be8add5eca973b/mii/batching/ragged_batching.py#L675) a `uid` for each request using the `_get_uid()` API. The `_get_uid()` function randomly generates an unique `uid` in the range of `[self.UID_RANGE_LB, self.UID_RANGE_UB)` before releasing the thread lock: https://github.com/microsoft/DeepSpeed-MII/blob/c171c4ee290e96c0d3e618b654be8add5eca973b/mii/batching/ragged_batching.py#L657-L664

This `uid` is removed from `self.uids` when the request is successfully completed in [`GeneratorReply`](https://github.com/microsoft/DeepSpeed-MII/blob/c171c4ee290e96c0d3e618b654be8add5eca973b/mii/grpc_related/modelresponse_server.py#L85) using the `MIIAsyncPipeline.flush_uid` method.

However, if any exception occurs in `MIIAsyncPipeline.put_request` after getting a new uid, or in `MIIAsyncPipeline.get_response` , the exception will terminate `GeneratorReply` and propagate upstream. The `uid` will never be removed from `self.uids` in this case. `UID_RANGE_LB` and `UID_RANGE_UB` are set to `1`, and `10000` in `MIIAsyncPipeline`, respectively. Therefore, after 9999 exceptions in `MIIAsyncPipeline.put_request` or `MIIAsyncPipeline.get_response`, there will be 9999 leftover uids in `self.uids` and no available choices for a new `uid`. Thus the following `while` block will never exit.
```python
            while uid in self.uids:
                uid = random.randrange(self.UID_RANGE_LB, self.UID_RANGE_UB)
```
[...]

PR #490, opened 2024-06-05
Always Flush UIDs after `GeneratorReply`

weiqisun:
This PR fixes the issue where the MII server hangs after encountering 9999 exceptions. This issue was initially reported in https://github.com/microsoft/DeepSpeed-MII/issues/489.


weiqisun:
NVM, this fix doesn't solve the issue since the `uid` is never returned

PR #492, opened 2024-06-07
Remove Conversation from MII as it was deprecated and removed from transformers.

loadams:
Fixes in MII that correspond to [these changes](https://github.com/microsoft/DeepSpeed/pull/5629) in DeepSpeed

### DIFF
--- a/mii/legacy/examples/local/conversational-example.py
+++ b/mii/legacy/examples/local/conversational-example.py
@@ -11,4 +11,4 @@ name = "microsoft/DialoGPT-large"
 
 print(f"Deploying {name}...")
 
-mii.deploy(task='conversational', model=name, deployment_name=name + "_deployment")
+mii.deploy(task='question-answering', model=name, deployment_name=name + "_deployment")
--- a/mii/legacy/examples/local/conversational-example.py
+++ b/mii/legacy/examples/local/conversational-example.py
@@ -11,4 +11,4 @@ name = "microsoft/DialoGPT-large"
 
 print(f"Deploying {name}...")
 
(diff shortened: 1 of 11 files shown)


adk9 (approved):

mrwyattii:
Need to remove the conversational task type unit test, set minimum version of transformers to whatever version removed the task type, and update docs/READMEs to reflect these changes

requirements/requirements.txt
```diff
@@ -10,6 +10,6 @@ pydantic
 pyzmq
 safetensors
 torch
-transformers
+transformers>=4.41.2
```
adk9, line 13: `ConversationalPipeline` and `Conversation` has been deprecated for a while and was removed in transformers 4.41.2. Do we really have to pin this version as minimum? These changes should be backward-compatible.

mii/legacy/examples/local/conversational-example.py
```diff
@@ -11,4 +11,4 @@
 
 print(f"Deploying {name}...")
 
-mii.deploy(task='conversational', model=name, deployment_name=name + "_deployment")
+mii.deploy(task='question-answering', model=name, deployment_name=name + "_deployment")
```
adk9, line 14: `TextGenerationPipeline` is recommended as replacement for `ConversationalPipeline`, so you likely want `task='text-generation'` here.

requirements/requirements.txt
loadams: Good point, unpinning.

mii/legacy/examples/local/conversational-example.py
loadams: Good catch, updated.

PR #512, opened 2024-07-25
Unpin transformers version

loadams:


mii/legacy/models/load_models.py
```diff
@@ -74,8 +74,8 @@ def load_models(model_config):
                                           config=inf_config)
         if model_config.profile_model_time:
             engine.profile_model_time()
-        if hasattr(inference_pipeline, "model"):
-            inference_pipeline.model = engine
+        #if hasattr(inference_pipeline, "model"):
+        #inference_pipeline.model = engine
```
loadams, line 78: It seems we lose the torch_dtype when replacing this, so we need to copy over the missing components first.

PR #556, opened 2025-02-25
Add pyproject.toml

loadams:


.github/workflows/formatting.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          grep -E "clang-format|pre-commit" requirements/requirements-dev.txt | xargs pip install
+          grep -E "clang-format|pre-commit" pyproject.toml | xargs pip install
```
loadams, line 31: Need to handle the readthedocs case too.

